PR #3, opened 2017-11-08
proto3 switch-over RFC

cschuet:
[Rendered](https://github.com/cschuet/rfcs/blob/proto3/text/0000-proto3-transition.md)

PR #2, opened 2017-11-03
Update readme and add RFC template.

cschuet:

PR #1, opened 2017-10-26
Update readme and add RFC template.

cschuet:

PR #4, opened 2017-11-09
Improved handling of multiple trajectories and bags in the offline node

ojura:
[Rendered](https://github.com/googlecartographer/rfcs/blob/master/text/0009-offline-node-improved-trajectories-bags.md)


text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,90 @@
...
+
+The offline node currently supports: __A1a__, __A/B 2c__.
+
+The online node can never support __B3__, because only single robot's data can be published on a topic at a time.
+Also, __B3__ and __a/b__ is impossible for the same reasons.
+
+Note that some combinations are invalid, such as __A3__ (a single robot can't be on multiple concurrent trajectories), and __B/C 1__ (there can't be a single trajectory if there are multiple robots).
+Furthermore, __A/B 2 a/b__ would require us to be able to specify when a trajectory stops and begins within the same bag aggregate and on the same range data topics, which would be too complicated and out of scope.
+
+The aim of the RFC is to enable the user to use the offline node for __C2/3__ like in the online node, to preserve the ability for __A/B 2c__, and to possibly offer additional functionality: for example, __b__, and __B3c__, which is otherwise impossible with the online node.
```
wohe: What I'm missing here is a true motivation, i.e. why are you interested in seeing trajectories being built at the same time in the offline node. Especially given that it has an unclear effect on the end result (all data will be processed whether in parallel or sequential) and it does not change the computational effort. For the main use case of the offline node right now, i.e. process data as quickly as possible and output the result, this seems irrelevant.

ojura: You are right about this part - being able to watch concurrent trajectories being built at the same time does not influence the end result in any way besides potentially increasing the coolness factor. However, it's possible to do it with the online node. Why not have the ability to do the same thing in the offline node, at increased processing speed? I believe that the offline node should offer everythihg that the online node can do, but at greater speed.

However, if I am correct, the offline node does not currently support concurrent trajectories _at all_. Or different robots.

The true motivation is stated at the beginning - to make the offline node more flexible and possible to be used in these cases. Someone who has bagged a swarm of robots and wants to try out cooperative SLAM with Cartographer can also be interested in crunching the bags as fast as possible. Perhaps they have a bag for each robot, perhaps they have all robots in a single bag - I think we can handle all these cases without much problems if we use Configuration #2 or #3.

wohe: I guess what you explained now is what I was missing. Add these use cases (concurrent trajectories being built for additional coolness, different robots) at the very beginning. This seems to be the motivation to me. The other discussion is already the way how to get there.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+  - Performing multiple calls of the `StartTrajectory` service, with possibly different `TrajectoryOptions` and necessarily different sensor data topics (__C__); all is specified in the service request.
+  - Alternatively, the `start_trajectory` helper node is invoked multiple times, with possibly different `.lua` files for `TrajectoryOptions` and necessarily different sensor data topics (__C__); the recently merged PR googlecartographer/cartographer_ros#584 enables specifying different topics via topic remapping.
+  - Specially, if a trajectory is finalized, we may start another independent trajectory which uses the same sensor data topics and possibly different `TrajectoryOptions` (__B2__).
+  This is already achievable in the offline node by simply loading multiple bags, which creates a trajectory for each bag, although `TrajectoryOptions` are necessarily the same for all trajectories (__B2c__). 
+  The ability to do this has to be preserved or expanded.
+
+If we suppose that there is a mandated 1:1 relationship between bags and trajectories in the offline node, the following configuration can be proposed.
+
+### Configuration #1
```
cschuet: I would advise to keep the design and scope limited to what is absolutely necessary to solve your usecase. Sure it would be nice to have the additional functionality of options 2 and 3, but with some bag surgery it seems you can reduce both of these to option 1. I would suggest to move ahead with that.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+
+__For each bag (trajectory)__, the configuration is a __tuple__ consisting of:
+  - `TrajectoryOptions` (given in a `.lua` file)
+  - A set of sensor data topics
+  - Boolean `use_bag_transforms` which allows us to ignore transforms in the bag
```
cschuet: How would this work with multiple bags setting use_bag_transforms to true? Couldn't the frame names clash and how do you want to deal with that?

ojura: Each bag (or `BagAggregate` in option 3) would have its own transform buffer, as specified in the RFC. This would enable you to track multiple concurrent robots with same frame names.

text/0000-offline-node-improved-trajectories-bags.md
SirVer: +1. All functionality needs to be taught somehow. Cartographer is already a complex system, so adding more features will make it even harder to understand. Multi trajectory is not a self evident concept in general, so making it as simple to understand as possible is important.

My suggestion: Move this explanation to discussion points, explain why handling one trajectory per bag is sufficient with some bag filtering and that we limited the scope to this, then simplify the rest of the rfc to only deal with this use case.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+[motivation]: #motivation
+
+The online node currently supports tracking multiple concurrent trajectories in real time.
+This enables the Cartographer user to do cooperative SLAM using multiple robots.
+However, support for this in the offline node is currently limited - only a single robot configuration may be used, and the trajectories cannot be concurrent.
+Since this functionality is supported in the online node, this RFC aims to expand the offline node to be able to provide the same functionality, but at increased speed. 
+
+For example, a user has bagged a swarm of robots and wants to try out cooperative SLAM with Cartographer.
+They are interested in crunching the bags as fast as possible.
+Perhaps they have a bag for each robot, perhaps they have all robots in a single bag - the offline node should have a flexible configuration interface to allow for various possibilities.
```
SirVer: I disagree with this still. Why should we support one bag - multiple robots? That is just an `rosbag filter` away from a sane input to cartographer_ros. It also deals better with the problem of how to replay TF messages in the bag - if you have one per robot, they are namespaced to the robot.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+
+The concurrent trajectories could be processed either sequentially or in parallel - it should not influence the end result in any way, since local SLAM is not influenced by other trajectories, and global SLAM should arrive at the same pose graph at the end, no matter what the order of insertion is.
+However, processing and displaying the concurrent trajectories in parallel, like the online node can do, would be more useful and increase the coolness factor.
```
SirVer: This could also be handled by a rosbag filter or a short python script. Essentially translate each `header.stamp` in the bag to a common start time for multiple bags. We could add a script into cartographer_ros to do this. No need for the offline node to support that I think.

SirVer: To clarify: that means that the offline node could just merge sort from multiple bags: always process the newest message and push it through the pipeline. It would still work exactly the same in the current use case (just relaxing that the bags must be provided in the same time order), but would automagically support playing bags with overlapping timestamps.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+### Overview of possible dataset scenarios
+Let's give an overview of all possibilities how a dataset layout and content could look like:
+
+| Robot plurality                                                         | Trajectory plurality                                 | Bag plurality                                                          |
```
SirVer: I think the goal here is to support `N bags, from N robots, with N configurations`. Everything else is a subset of this. We should not care if the data overlaps or not - as said, that can be changed rewriting the bags data.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+This is given in Configuration #2.
+
+### Configuration #2
+
+Similar to Configuration #1, but have a __set of tuples for each bag__.
+Each tuple describes a trajectory in the bag.
+
+For both Configuration #1 and Configuration #2, all concurrent trajectories (whether in the same bag as permitted by Configuration #2, or in different bags) are __collated__.
+
+### Handling split bags
```
SirVer: Joining bags can be done by a script too. Let's not rebuild features that ROS tools already give us. I am fine with adding some scripts (to be outlined here) to cartographer_ros that help with joining and reshuffling timestamps for bags. This will also allow to drop the concept of a BagAggregate - which seems an unnecessary complexity in this RFC.

text/0000-offline-node-improved-trajectories-bags.md
```diff
@@ -0,0 +1,106 @@
...
+
+This could be handled by defining a `BagAggregate`: a list of one or more continuous bags, which are to be treated as one bag.
+
+### Configuration #3
+Configuration is given as a __set of tuples for each `BagAggregate`__, each tuple describing a trajectory in the `BagAggregate`.
+
+As before, any concurrent trajectories are collated.
+Each `BagAggregate` has its own transform buffer (which is shared by all bags in the aggregate).
+
+### Data format of the configuration
```
SirVer: This section is most interesting to me - but it needs to be written. For example would we replace the current command line options or add a new one that reads a lua file as input? How would such a Lua file look?

text/0000-offline-node-improved-trajectories-bags.md
ojura: > To clarify: that means that the offline node could just merge sort from multiple bags: always process the newest message and push it through the pipeline. It would still work exactly the same in the current use case (just relaxing that the bags must be provided in the same time order), but would automagically support playing bags with overlapping timestamps.

@SirVer That's right! So there is no need to remap time in the bags, right?

I agree with you on the rest (user can split one bag into multiple bags, or join multiple bags into one, so that in the end, the data obeys 1 trajectory : 1 bag format). I think you misunderstood the text in the RFC: the user should never remap time in the bags. The offline node should accept both simultaneous and sequential trajectories/bags. 

@wohe suggested that we could process simultaneous trajectories sequentially, to which I responded that it would be better if we processed them truly simultaneously like you described, by merge sorting them (I called this collating, was that wrong? Better change it to merge-sort.).

> process the newest message and push it through the pipeline
 
You mean the oldest? :-)

text/0000-offline-node-improved-trajectories-bags.md
ojura: #helpneeded :D I am currently a bit short on time, but I'll get back to this RFC when I can.

text/0000-offline-node-improved-trajectories-bags.md
SirVer: > @SirVer That's right! So there is no need to remap time in the bags, right?

Only if you want to have the bags from one robot (i.e. the deutsches museum runs) looking like they were recorded at the same time. You know, for the coolness factor.

> I called this collating, was that wrong? Better change it to merge-sort.).

I think collating is correct as well - merge sorting is the means to get collated data here. 

> You mean the oldest? :-)

yes, sure.

Will you update the RFC with these new information?

ojura:
Hello everyone, and happy holidays! @SirVer I finally got around to making some progress on this.

The implementation which follows the proposed configuration #1 in the rfc has been implemented in this branch: https://github.com/googlecartographer/cartographer_ros/compare/master...larics:simultaneous-bags (PR googlecartographer/cartographer_ros#636)

I have expanded the comma-separated list approach, as already used for `bag_filenames`, to `configuration_basenames`, `urdf_filenames`, and there is now `sensor_topics` as well.

Here is an example of launching the offline node for two robots named alpha and delta, which were driven simultaneously:

```
rosrun cartographer_ros cartographer_offline_node
-configuration_directory /somewhere
-configuration_basenames alpha.lua,delta.lua -urdf_filenames alpha.urdf,delta.urdf
-bag_filenames alpha.bag,delta.bag -sensor_topics alpha/scan:alpha/pose,delta/scan:delta/pose
```

Note how sensor topics for different trajectories are separated by commas, and further separated with colons. Leaving sensor topics unspecified causes the default computed topic names to be used.

If only one configuration basename or set of topics is specified, it will be used for all trajectories.

ojura:
I have just realized that multiple trajectories in single bags are supported by this as well. This can be achieved by simply specifying the same bag multiple times, but with different sets of sensor topics.

cschuet:
@ojura Thanks for the PR. I looked at it and the implemented features make sense to me. Be aware though that we will introduce a gRPC-ified version of the `cartographer_offline_node` in next days and your changes should also be added there. In essence this will mean moving most of your (then shared) code out of `offline_node_main.cc` into some other place so that both `offline_node_main.cc` and `offline_node_grpc_main.cc` can use it.

But before that, could you update the RFC with the most recent approach and move all other options into a "Alternatives considered" section or something like this? We should get the RFC approved first before discussing implementation in detail.

ojura:
Okay, I updated the rfc to be more in-line with the discussion and the PR.

cschuet:
@ojura Thanks! Can you address open comments by @SirVer which is asking to add certain information into this RFC? Otherwise this RFC LGTM.

@SirVer @wohe Can you take a look at this RFC again? Are you ok with moving forward on this? Tentative implementation is in googlecartographer/cartographer_ros#636

ojura:
@cschuet No problem. I think I covered @SirVer's request for adding data format description in the section `Data format of the configuration`. Do you think something is missing?

SirVer:
@ojura @cschuet  I am fine with moving forward.

cschuet (approved):

cschuet:
@wohe @SirVer Can you formally approve this. Then I will assign an RFC ID and merge it.

SirVer (approved):

cschuet:
Assigned RFC=0009

PR #5, opened 2017-11-10
RFC to rename sparse pose graph.

wohe:
See googlecartographer/cartographer#646.


text/0000-renaming-sparse-pose-graph.md
```diff
@@ -0,0 +1,39 @@
+# Renaming the Sparse Pose Graph
+
+## Summary
+[summary]: #summary
+
+The goal of this change is to improve readability of the code base by choosing
```
cschuet: one sentence per line.

cschuet (commented):
So I don't have a strong feeling about the rename. The only thing I am concerned about is that this is non-backwards compatible breaking change to our LUA files. If we communicate this well so that people that are broken can easily find out how to unbreak themselves I will not stand in the way of this rename.

SirVer:
Add that we can still do this before 1.0, but not after really. So we do it now.

gaschler:
I agree that the rename has a slight benefit and breaking api and config files now is still acceptable.

SirVer (approved):

gaschler (approved):

cschuet (approved):

text/0000-renaming-sparse-pose-graph.md
```diff
@@ -0,0 +1,39 @@
...
+Sparse Pose Adjustment is an algorithm for optimizing a pose graph using a direct linear solver and inspired the current implementation in Cartographer.
+This is an implementation detail and might change in the future.
+Having a name that sounds as if it refers to the current inspiration of the implementation thus is an argument for changing the name.
+
+### Isn't a sparse pose a well known concept in SLAM?
+
+No.
+The name refers to the sparsity of the graph not the pose.
+Removing "sparse" also gets rid of this misconception.
+A pose graph is a well known concept in SLAM, e.g. in the Sparse Pose Adjustment paper, 8 of the 9 mentions of a pose graph do not qualify it as "sparse".
```
ojura: Subjectively, I like the name so I'll be a bit sad to see it go. Although "sparse pose graph" seems to be a known concept in SLAM: a [search](https://www.google.com/search?q="sparse+pose+graph") brings up quite a few SLAM related-articles containing the term.

SirVer: Well, [pose graph SLAM](https://www.google.de/search?q=pose+graph+SLAM&oq=pose+grap&aqs=chrome.0.69i59j69i61j69i57j69i61l2j0.1079j0j1&sourceid=chrome&ie=UTF-8) returns much more. Looking through the first few of these hits, they are all sparse graphs. No need to carry this word around.

PR #6, opened 2017-11-10
Refactor odometry in SPG optimization problem

ojura:
[Rendered](https://github.com/larics/rfcs/blob/odometry-refactoring/text/0000-odometry-refactoring.md)

### DIFF
--- a/text/0000-odometry-refactoring.md
+++ b/text/0000-odometry-refactoring.md
@@ -1,0 +1,28 @@
+# Refactor odometry in SPG optimization problem
+
+## Summary
+[summary]: #summary
+
+Do not use `std::vector<TransformInterpolationBuffer>` in SPG optimization problem anymore.
+
+## Motivation
+[motivation]: #motivation
+
+`TransformInterpolationBuffer` does not support trimming-in-middle.
+
+Performing refactoring of odometry data is a pre-requisite for serialization/deserialization.
+
+## Approach
+[approach]: #approach
+
+Replace `std::vector<TransformInterpolationBuffer>` by `MapByTime<sensor::OdometryData>`?
+
+
+## Discussion Points
+[discussion]: #discussion
+
+How to implement the interpolation functionality?
+
+Expand `MapByTime`, or subclass it (`InterpolatedMapByTime`)?
+
+Where to use `transform::StampedTransform`, and where to use `sensor::OdometryData`?


text/0000-odometry-refactoring.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Summary
+[summary]: #summary
+
+Do not use `std::vector<TransformInterpolationBuffer>` in SPG optimization problem anymore.
+
+## Motivation
+[motivation]: #motivation
+
+`TransformInterpolationBuffer` does not support trimming-in-middle.
```
wohe, line 11: I think this RFC is too fine grained and this is also not the true motivation: why do we want trimming in the middle in the first place? I agree that we should have an RFC for https://github.com/googlecartographer/cartographer/projects/1 which motivates it. @cschuet Opinion?

ojura: You mentioned yesterday that you want to switch away from `TransformInterpolationBuffer` and follow how IMU data is now kept in `MapByTime`.

wohe: "Motivation" should be what caused us to get interested in this change. Trimming-in-the-middle is not a goal by itself, but a necessity to support life-long mapping.

ojura: I see. If you want, I can close this PR since it's too fine grained for an RFC and open an issue.

wohe: No need for an issue. We have a project already (see above) which predates the RFC process. We could add an RFC for it. Other than that, we will just do it.

ojura: The Projects UI is pretty much read-only for non-maintainers. Where will discussion take place?

> Other than that, we will just do it.

You mean you will do the refactoring? :-)

wohe: Yes. Already on it.

SirVer:
@ojura Can this be closed? It seems like it is about a refactoring that did not affect the public API. 
@wohe Opinions?

wohe:
I think this should be closed. The specific change it mentioned is implemented now, and if we would like to have an RFC for life-long mapping we should have a separate PR for it.

ojura:
For the future, should I just open issues for minor stuff like this?

wohe:
Not in my opinion. There should be RFCs on a higher level explaining the motivation for the change. The change itself should then be done in a PR. No need to make the process heavier. @cschuet What do you think?

PR #7, opened 2017-11-16
Cloud-based mapping RFC.

cschuet:
[Rendered](https://github.com/cschuet/rfcs/blob/3d76b6ba938d18adeae8cd08fc33c52134fd3bb1/text/0002-cloud-based-mapping-1.md)


text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,135 @@
...
+
+In the rest of this RFC we refer to the Cartographer instance running in the cloud as *Cloud Cartographer* and the instances running on the agents as *Agent Cartographer*.
+With the cloud only solving the global SLAM problem and the agents running in *pure localization mode*, we know that
+
+- *Cloud Cartographer* does not require a `LocalTrajectoryBuilder` as there is no local SLAM problem to be solved.
+- *Agent Cartographer* needs to upload all necessary information for *Cloud Cartographer* to formulate the global SLAM problem *and* also route that information to its local version of the global optimization problem that is reduced to the pure localization case.
+
+### Software Components Overview
+
+While implementing this RFC we want to keep the following three types of Cartgrapher users happy
```
wohe: typo: Cartographer

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Users that link against libcartographer as a C++ library and don’t care about ROS at all
+- ROS users that use the ROS binaries cartographer_offline_node and cartographer_node to stream sensor data as ROS messages and receive Cartographer’s SLAM results as ROS messages
```
wohe: Consistency in styling:
```
*cartographer_offline_node* and *cartographer_node*
```
?

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,135 @@
...
+This order needs to be maintained per sensor when sensor data is forwarded to the cloud.
+A simple way to achieve this is to use a streaming gRPC connection, which ensures that “each side will always get the other’s messages in the order they were written”.
+We choose to forward the sensor data in a streaming fashion rather than batching them up into a state that includes the finished submap for two reasons
+
+1. gRPC prefers small message sizes. Streaming does its best to ensure that.
+2. Continuously sending up sensor information distributes the computational load in the cloud better as it allows to search for INTER_SUBMAP constraints against scans that belong to yet unfinished submaps.
+
+#### Visualization of SLAM results
+
+The current state of affairs is documented in the appendix section “Visualization of SLAM results”.
```
wohe: There is no appendix. Remove this sentence?

wohe (approved):
Only nits.

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+In many robotics applications multiple robots simultaneously map and localize within a continuously changing environment.
+
+Forwarding local SLAM results from agents to a central instance and solving the global SLAM problem in the cloud has the following advantages
+
+- by involving multiple robots sharing information (e.g. sensor measurements or intermediate mapping results) among one another, local world observations can be combined into a global and more detailed map of the world and the mapping process expedited.
+- by moving some of the computational load of mapping away from battery-powered agents to a central cloud instance with virtually unbounded computational resources, the mapping performance and runtime of agents can be enhanced with reduced hardware requirements.
+
+
+## Approach
```
SirVer: Not sure if it belongs here, but what about scaling? Cloud cartographer can't solve significantly larger problems than we currently solve on the robots without some changes. Should this be mentioned here?

SirVer: Probably something for the last section.

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+- extract interfaces out of `MapBuilderBridge` and `SensorBridge`.
+- implement a gRPC bridging version of the above two classes that serialize any sensor data to proto3 and send that to a gRPC service.
+
+In principle once we have a binary that does ROS to gRPC translation it would be possible to replace the existing cartographer ROS node with a combination of this translation binary and *cartographer_grpc*, i.e. instead of the C++ interface between *cartographer_node* and libcartographer that exists today we would introduce a local, loopback gRPC layer.
+We have decided not to go this route and keep the current *cartographer_node* as is until we have a better idea of the performance impact (or benefit) of this additional gRPC serialization / deserialization and whether adding gRPC to our dependencies is acceptable for our ROS users.
+
+#### cartographer_grpc
+
+We introduce a new binary,  *cartographer_grpc*, in the *cartographer* repo, which offers the cartographer gRPC service - essentially a (flattened) gRPC representation of the `MapBuilder` interface.
```
SirVer: Was do you mean with flattened?

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+
+This will require us to
+
+- extract interfaces out of `MapBuilderBridge` and `SensorBridge`.
+- implement a gRPC bridging version of the above two classes that serialize any sensor data to proto3 and send that to a gRPC service.
+
+In principle once we have a binary that does ROS to gRPC translation it would be possible to replace the existing cartographer ROS node with a combination of this translation binary and *cartographer_grpc*, i.e. instead of the C++ interface between *cartographer_node* and libcartographer that exists today we would introduce a local, loopback gRPC layer.
+We have decided not to go this route and keep the current *cartographer_node* as is until we have a better idea of the performance impact (or benefit) of this additional gRPC serialization / deserialization and whether adding gRPC to our dependencies is acceptable for our ROS users.
+
+#### cartographer_grpc
```
SirVer: This is quite hand wavy. Will there be a follow RFC that explains that thing in more detail?

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+| cartographer_(offline)_node      | cartographer_ros | proto3, ROS       |
+| cartographer_(offline)_node_grpc | cartographer_ros | proto3, gRPC, ROS |
+
+### Detailed Design
+
+#### cartographer_grpc
+
+The important interaction between local and global SLAM happens through `GlobalTrajectoryBuilder` between `LocalTrajectoryBuilder` and `PoseGraph`. 
+
+The following component diagram describes the cloud mapping architecture
```
SirVer: I guess the right side is "Cloud Cartographer" while the left one is "Agent Cartographer". Might make sense to point that out.

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+i) can be detected by `GrpcMapBuilderBridge` remembering the last submap it has seen as `insertion_result.back()`.
+Whenever that changes between invocations of `AddRangeFinderData()`, a new submap has been added.
+`GrpcMapBuilderBridge` then sends up the `local_pose` of the new submap to the cloud.
+
+ii) can be detected by `GrpcMapBuilderBridge` observing `insertion_submaps.front().finished()`.
+Whenever that returns true (which is only once throughout the lifetime of any one `Submap`) the submap’s `Probability-` or `HybridGrid` is serialized and uploaded to the cloud.
+
+The `CollatedTrajectoryBuilder` feeds sensor data into the `GlobalTrajectoryBuilder` in a strict time order.
+This order needs to be maintained per sensor when sensor data is forwarded to the cloud.
+A simple way to achieve this is to use a streaming gRPC connection, which ensures that “each side will always get the other’s messages in the order they were written”.
```
SirVer: link the page with the quote?

text/0000-cloud-based-mapping.md
```diff
@@ -0,0 +1,134 @@
...
+We choose to forward the sensor data in a streaming fashion rather than batching them up into a state that includes the finished submap for two reasons
+
+1. gRPC prefers small message sizes. Streaming does its best to ensure that.
```
SirVer: Mention somewhere that we ignore the intermittent connection dropouts that are likely to happen in real systems and address them later somehow.

jihoonl:
I would like to add some comments to describe our use case in a couple of days. When do you plan to merge this rfc?

cschuet:
@jihoonl I will wait for your comments then.

jihoonl (changes requested):
This rfc clearly states the plan towards cloud mapping support. Thank you for preparation @cschuet!

I have a couple of questions here.

1.
This image is what I understood how data flows in cartographer. *Cloud cartographer* receives `LocalSlamResult`(+ odom, imu and etc) and computes global slam(loop closure constraint building + optimization). In the meantime, *Agent cartographer* will split into two processes. `cartographer_grpc` receives grpc format raw data, generates `LocalSlamResult`, forwards it to Cloud and `cartographer_node_grpc`. And `cartographer_node_grpc` is a bridge node that converts ros msgs to protobuf.

cloud_carto

I would like to add another node in *Agent Cartographer* that communicates with *libcartographer* using standard C++ interface but still forwards data to cloud via grpc. Having `cartographer_node_grpc` as a bridge node seems to create extra memory coping step for me. Since *cartographer* deals with multiple of large size data format(especially point cloud), I would like to avoid inter-process data transfer as much as possible. For example, nodelet [[1](http://wiki.ros.org/nodelet)], [[2](http://www.clearpathrobotics.com/assets/guides/ros/Nodelet%20Everything.html)] is used  in ROS for high-frequency and bandwidth data. Also, it would be a cleaner structure if all internal-robot communication is kept in ROS format while grpc is used for all external-robot communication.

How do you think? I think having two interfaces, one for robot and one for cloud, in `MapBuilder` isn't breaking the current structure. One way of implementing it I can think of is allowing `MapBuilder` to have an array of `local_slam_result_callback`. Then, `MapBuilderBridge` in cartographer_ros gets its result just as now, and grpc based *Cloud Bridge* can also communicate with cartographer in cloud.

2. 
In the current RFC, I see that *Agent Cartographer* uploads `LocalSlamResult` to cloud for global SLAM. But, how does it downloads intermediate global SLAM results from *Cloud Cartographer*? Since robots localizes against its “local” pose_graph, I think it should receive updates from cloud.

3. 
How would it work in a completely new environment? In principle, it is clear that robot runs in pure localization mode and uploads its local result to cloud for global mapping. But, I am wondering how it would work when robots should create a map from scratch. Could you elaborate this use case?

4. 
Assume cartographer is already running in the cloud and you would like to start the robot in pure localization mode, how would it receive(or initialize) the pose graph from cloud when it starts?

5. 
[Cartographer_grpc offers SLAM solutoin as gRPC message](https://github.com/cschuet/rfcs/blame/cloud-mapping/text/0000-cloud-based-mapping.md#L67). Does it mean that there will be a light-weighted node to monitor global SLAM result in remote? :)

text/0000-cloud-based-mapping.md
cschuet: Good point. Life-long mapping which aims to make the optimization problem feasible when robots continuously move in a constrained area for an arbitrary long time and some from of sharding in real space for huge mapping problems is not addressed by this RFC. I  will specifically call this out in one of the discussion points.

text/0000-cloud-based-mapping.md
cschuet: Yes. The purpose of this RFC is to agree on the general direction. It is also quite long already. Specific questions like how do we handle connection break-ups and such will be dealt with in break-out RFCs.

text/0000-cloud-based-mapping.md
cschuet: Sorry this is confusing and that's why I will remove "flattened". What I mean was that MapBuilder currenly has methods which return complex types, e.g. SensorBridge, that are used to feed data. This won't work for RFCs. With flattened I meant that

~~~
SensorBridge* sensor_bridge = map_builder_bridge->GetSensorBridge(trajectory_id);
sensor_bridge->AddImuData(...);
~~~
will become one RPC which adds IMU data for a specific trajectory.

text/0000-cloud-based-mapping.md
cschuet: Thanks. Done.

text/0000-cloud-based-mapping.md
cschuet: Done.

text/0000-cloud-based-mapping.md
cschuet: Done.

cschuet:
1. I see your point about extra serialization-deserialization. We were also discussing this and had the suspicion that moving the ROS deserialization (which is quite costly) to its own process might actually be a good thing separating that from the local SLAM thread. What you seem to describe is essentially the `cartographer_node` with added gRPC upload capabilities. In order to keep pure ROS users happy that are not interested in gRPC we decided to keep `cartographer_node` free of gRPC for now. So I would suggest let's start with the current proposal and if performance is not good, we will solve the problem by adding gRPC support to `cartographer_node` or introducing a new binary as you proposed.

2. You are right in this RFC the back-channel for propagating global SLAM results from the Cloud to robots is not described. The simplest version would be an RFC in the Cartographer cloud service which allows robots to download a "pbstream" of the latest global map. I have deliberately left this for a separate RFC as I think the right way to tackle this is an interesting discussion on its own.

3. `cartographer_grpc` will be very configurable. Whether you forward local SLAM results or not you can decided and what version of the global problem (pure localization or full problem) you want to solve on the robot, you can decide independently of one another.

4. See 2). My favourite option would be an RPC where you can get the "pbstream" but let's discuss this in another RFC.

5. Yes. We were already discussing with our larger team that there might be an opportunity to build web-based visualization of mapping results that you can host in the cloud. So stay tuned :)

jihoonl:
Thank you for the explanation. @cschuet. Could you add a note that 2 and 4 will be addressed in another rfc? 

And could you rename RFC to something like *cartographer_grpc design for cloud-based mapping* or *Cloud-based Mapping RFC Part 1*? Cloud-based Mapping seems too broad to describe its actual scope.

> Yes. We were already discussing with our larger team that there might be an opportunity to build web-based visualization of mapping results that you can host in the cloud. So stay tuned :)

Awesome!

cschuet:
Done.

jihoonl (approved):

cschuet:
@SirVer I hear you are on vacation, so I will go ahead and merge the RFC. I addressed all your comments. Enjoy the time off!

@wally-the-cartographer merge

cschuet:
@jihoonl I just realized that gRPC even supports unix domain sockets ([see here](https://grpc.io/grpc/cpp/md_doc_naming.html)) as an IPC mechanism, so I am really hopeful that running 
`cartographer_node_grpc` and `cartographer_grpc` is feasible.

jihoonl:
@cschuet The thing is that hundreds of processes are already running in the robot for localization, navigation, manipulation, and other tasks. And this is why I need to and try to keep every component as minimum as possible. Other developers complained already that cartographer consumes significantly larger portion of CPU and memory than previous localization method(static map + MCL). Permanent CPU consumption for localization has already increased from 3% to 10%.

I understand that it was inevitable increase to use cartographer because it provides better performance and dynamic environment adaptability. But, `cartographer_grpc` and `cartogapher_node_grpc` modulization seem to introduce just an additional IPC to do the same task. This is why I wanted to avoid. 
Anyway, I also hope that it is too much concern. It might have almost no overhead. Let's see if it is feasible first.

PR #9, opened 2017-11-20
Reconsider the role of odometry in the 2D optimization problem

MichaelGrupp:
[Rendered version](https://github.com/MichaelGrupp/rfcs/blob/master/text/0000-odometry-use-in-optimization-problem.md)


text/0000-odometry-use-in-optimization-problem.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Approach
+[approach]: #approach
+
+We didn't come up with a master plan for this yet, but two possible approaches are:
+
+1. Revert the changes introduced by the PR mentioned above if the discussion yields that the odometry constraint is *generally* not beneficial.
+
+2. A simple compromise would be to introduce a boolean option that allows to use the initial pose estimates for the relative constraints instead of the odometry even if odometry data is available (as it was before the change).
+
```
Mofef: Another option would be to use both the local slam result as well as the wheel odometry, each weighted by a individual set of translational and rotational weight parameter.
This way everyone could adapt cartographer to his sensors.
Previous discussions suggest that for some setups the mentioned pull request yielded an improvement, for some others it decreased quality.

MichaelGrupp: Added.

text/0000-odometry-use-in-optimization-problem.md
```diff
@@ -0,0 +1,37 @@
...
+* Odometry-based constraints imply a certain level of trust in the quality of the odometry, which could be problematic due to the usually noisy nature of (wheel) odometry.
+* Initial pose estimation in the front-end combines different sensor modalities (pose extrapolation, scan matching) - in contrast, adding a pure odometry constraint in the back-end does not account for multiple sensors and may introduce a bias.
+
+## Approach
+[approach]: #approach
+
+We didn't come up with a master plan for this yet, but two possible approaches are:
+
+1. Revert the changes introduced by the PR mentioned above if the discussion yields that the odometry constraint is *generally* not beneficial.
```
cschuet: We are using odometry very successfully with our Fetch robots so I think it is safe to say that odometry is not *generally* non-beneficial.

text/0000-odometry-use-in-optimization-problem.md
```diff
@@ -0,0 +1,37 @@
...
+
+2. A simple compromise would be to introduce a boolean option that allows to use the initial pose estimates for the relative constraints instead of the odometry even if odometry data is available (as it was before the change).
+
+3. Use both the local slam result as well as the wheel odometry, each weighted by a individual set of translational and rotational weight parameter.
```
cschuet: I think generally 3 is preferred over 2, but generally I would like to understand in what the cases you have a noisy odometry which you don't want to give up completely because it seems to be useful for your local SLAM but disadvantageous for global SLAM.

text/0000-odometry-use-in-optimization-problem.md
```diff
@@ -0,0 +1,37 @@
...
+This way everyone could adapt cartographer to his sensors.
+
+We implemented 2. for testing and saw a qualitative improvement with our system.
```
cschuet: Can you elaborate a little more your setup? Did you ensure that you can't achieve the same effect by either not providing odometry at all and relying on the correlative scan matcher (are you using that already?) or just tuning down the weight of odometry int the global problem.

cschuet (commented):
Thanks for the RFC! See my comments.

text/0000-odometry-use-in-optimization-problem.md
Mofef: We are using the CorrelativeSM, yes. If we disable odometry all together we need to increase the search window of the CorrelativeSM, resulting in higher CPU load. 
Even if tuning down `consecutive_node_rotation/translation_weight` (is that what you mean by weight of odometry?) results in better results for one test case it makes the algorithm generally less stable.

MichaelGrupp:
As briefly mentioned in the Open House yesterday, I added the idea of using the pose extrapolator (or something similar) for the constraint.

text/0000-odometry-use-in-optimization-problem.md
MichaelGrupp: See the new point 4. The benefit of odometry in the local SLAM (i.e. pose extrapolation) is to provide a translational motion measurement, while the rotation part is extrapolated with the IMU.

MichaelGrupp (commented):
After internal discussion, @Mofef, @jihoonl and me now agreed that it makes sense to include a "motion model" constraint in the global optimization that is *not* based on / including the scan matching result.

However, we propose to use the pose extrapolator instead of pure odometry for that purpose as it combines the complementary strengths of IMU (rotation) and odometry (translation). See also point 4. in the proposal.

What do you think of this proposal? Is it worth giving it a try or do you have concerns that we might have to consider?

gaschler:
I also have the intuition that point 4 (global cost term using combined odometry/IMU estimates from PoseExtrapolator) is worth trying an experiment, I'm looking forward to your results.
However, I'm afraid that

- integration from PoseExtrapolator to PoseGraph is not straightforward, so I'd like to see a good approach
- we shouldn't diverge from the 3D pipeline, please include 3D in your experiments and approach.

MichaelGrupp:
Thanks for the feedback, we'll keep you updated once we have results for 2D and 3D (I have other more urgent things on the to-do list, so don't expect much in the next days).

I agree that an elegant approach for passing the `PoseExtrapolator` result to the `PoseGraph` is not straight-forward. For experimenting, I just hacked the pose extrapolation into the `MatchingResult` of the `LocalTrajectoryBuilder` to get it into the global scope. If the results show that the PoseExtrapolator is beneficial, we can think of a cleaner approach.

MichaelGrupp:
I finally updated the proposal with a (very brief) update of the outcome of our experiments that showed that the `PoseExtrapolator`is *not* beneficial in global optimization. More detailed results are available upon request.

As discussed in our offline meeting, we can try separately weighted constraints for `initial_pose` and `odom`. I'll do a test with that and keep you updated.

cschuet:
@MichaelGrupp  Any updates on separately weighted constraints for `initial_pose` and `odom`?

MichaelGrupp:
Implemented a prototype before christmas holidays and it *works*, but more in-depth tests are pending (time...) .

The way I implemented it at the moment is:
* the `initial_pose` constraint is always inserted
* a second constraint based on `odom` is inserted if odometry data is available
* both constraints can be weighted with the 4 parameters:
  * `initial_pose_{translation, rotation}_weight`
  * `odometry_{translation, rotation}_weight`
* these 4 parameters replace `consecutive_node_{translation, rotation}_weight`

Related commit on experimental branch: https://github.com/magazino/cartographer/commit/62eadefebe6c15c2f7fb9e1ca12f0cc9ac88aeef

Corresponding lines in `optimization_problem.cc` (2D): [see here](https://github.com/magazino/cartographer/blob/62eadefebe6c15c2f7fb9e1ca12f0cc9ac88aeef/cartographer/mapping_2d/pose_graph/optimization_problem.cc#L186)

Feel free to add your thoughts.

MichaelGrupp:
@cschuet To finally revive this dinosaur PR: I had a look at this again and played with our `cartographer_magazino` integration. The prototype mentioned above was also rebased onto a more recent upstream version (march 9th), see this branch: https://github.com/magazino/cartographer/tree/separate_relative_pose_constraints

My verdict is that the approach implemented in the prototype works well and the additional parameter flexibility adds a very good value when tuning. For example, I could down-weight the wheel odometry's rotation weight for the TORU datasets because it drifts especially during rotations. As opposed to the current `master`, I can still benefit from the reliable rotation estimates of the scan matcher due to the separate constraints. For example, I used:
```lua
POSE_GRAPH.optimization_problem.initial_pose_translation_weight = 1e5
POSE_GRAPH.optimization_problem.initial_pose_rotation_weight = 1e5
POSE_GRAPH.optimization_problem.odometry_translation_weight = 1e5
POSE_GRAPH.optimization_problem.odometry_rotation_weight = 1e1
```
I also had a look at the trajectories and couldn't see any smoothness or instabilty issues.

I don't have a dedicated report, but here's an example that shows a big improvement of the point-cloud in a problematic section when running the offline node (mapping) with the `hallway_localization.bag` in `highres` mode. The parameters shown above were the only change.
With current `upstream`:
upstream
With my prototype of separate constraints:
separate

I don't have tests on 3D data with odometry available, although the approach should work there as well.

MichaelGrupp:
@cschuet @gaschler Added the results into the RFC document, it's ready to be merged.

Just to check back before doing a code pull request in `cartographer`: 

Rename `initial_pose_{rotation, translation}_weight` to `local_slam_pose_{rotation, translation}_weight`?

cschuet (approved):

gaschler (approved):

PR #10, opened 2017-11-24
Serve ROS map from a pbstream file

Mofef:
[Rendered](https://github.com/Mofef/rfcs/blob/serve-ros-map-from-pbstream/text/0000-serve-ros-map-from-pbstream.md)

### DIFF
--- a/text/0000-serve-ros-map-from-pbstream.md
+++ b/text/0000-serve-ros-map-from-pbstream.md
@@ -16,11 +16,7 @@ If the pbstream gets frequently updated re-generating a `.pgm` and `.yaml` file 
 ## Approach
 [approach]: #approach
 
-This feature could either be implemented by modifying `occupancy_grid_node` to accept an alternative input as a pbstream file,
-or by modifying `pbstream_to_ros_map` to publish a rostopic if a flag is set. Conceptually the first option is less intrusive I would argue.
-Either way this probably may require some refacturing of one or both those nodes code, so it could be reused.
+This feature will live in a new binary along with `occupancy_grid_node` and `pbstream_to_ros_map`. Existing code for deserializing protos from a pbstream file will be reused by factoring it out of `pbstream_to_ros_map` and bringing it to libcartographer.
 
 ## Discussion Points
 [discussion]: #discussion
-
-* Whether to modify `occupancy_grid_node` or `pbstream_to_ros_map` or even to introduce another binary.
--- a/text/0000-serve-ros-map-from-pbstream.md
+++ b/text/0000-serve-ros-map-from-pbstream.md
@@ -1,0 +1,26 @@
+# Serve ROS map from a pbstream file
+
+## Summary
+[summary]: #summary
+
+We would like to have a ROS node that serves a `nav_msgs/OccupancyGrid` message like the `occupancy_grid_node` does.
+But instead of publishing the current state of cartographer, a frozen state from a pbstream file should be read and compiled into a flat map, which is then published.
+
+
+## Motivation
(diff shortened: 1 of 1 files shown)


text/0000-serve-ros-map-from-pbstream.md
```diff
@@ -0,0 +1,26 @@
...
+
+We would like to have a ROS node that serves a `nav_msgs/OccupancyGrid` message like the `occupancy_grid_node` does.
+But instead of publishing the current state of cartographer, a frozen state from a pbstream file should be read and compiled into a flat map, which is then published.
+
+
+## Motivation
+[motivation]: #motivation
+
+Since `OccupancyGrid` is the standard message type for map, many nodes depend on such a topic.
+If the pbstream gets frequently updated re-generating a `.pgm` and `.yaml` file in oder to read it in again using a ROS `map_server` is administrative overhead and generates redundant data.
```
cschuet, line 14: If the pbstream is often updated, how would this be solved if you had this binary which publishes the map as a topic? Would this binary be watching the file? Would you need to restart the binary?

Mofef: Good question. We were actually just thinking of a static map server that reads the pbstream file once. But maybe a watch_and_update command line flag would be worth thinking about. 
Do you have a use case in mind that is not already covered by the occupancy grid node?

gaschler:
(To document my two cents:)
This feature does not require any upstream changes to cartographer(_ros), so it could live in a different repo without maintenance effort.
So the question is whether it is such a frequently used tool that should be located in the main repo for users to find it.
I am tempted to say that reading a pbstream file is not the most common setup to publish a map.
If a very small and "obviously correct" implementation (easy to maintain) is possible, then, from my side, it could go into this repo.

Let's wait for at least another comment.

jihoonl:
I have created a proposed change in googlecartographer/cartographer_ros#622

The main reason is that we do not want to maintain additional environment representation files since pbstream file has everything we need. In our usecase, we update existing pbstream file once a week(only places where robot loses localization due to the large change in environment) and runs other localization methods([amcl](http://wiki.ros.org/amcl) or [mrpt](http://wiki.ros.org/mrpt_localization)) for some reasons(e.g increase of memory consumption). To do this so far, we had to 1) run cartographer slam, 2) generate pbstream file, 3) run pbstream_to_ros_map, 4) run map_server with created map file. And I think the step 3 is unhandy and not necessary.

`cartographer_ros` currently supports online occupancy grid feeding directly from `cartographer` and occupancy grid file generation from pbstream file. And this RFC is to support online occupancy grid feeding from pbstream file. I think `pbstream_to_ros_map` was introduced to reduce the complexity of map exporting process. I would like to introduce this for the similar reason.

Consider this RFC as an effort to extend the usability of `cartographer_ros` with a little change. It will be helpful to mix use of `cartographer` and other ros components.

SirVer:
I think this is already supported: Launch the node with loading the map and do not feed it the data. Also launch the `occupancy_grid_node` -> voila. 

This is also fairly efficient - cartographer will not do anything until more data is presented and the `occupancy_grid_node` will only generate the published map exactly once. Is that not sufficient?

jihoonl:
I ran a quick test and confirmed that the `cartographer_node` consumes 3.5% cpu in average with no trajectory presented whereas static map loading from pbstream consumes 0.7% cpu in average. If `cartographer_node` can consume less than 1% cpu, then I'm ok with the proposed solution.

jihoonl:
@cschuet can you review this rfc again?

cschuet (approved):

PR #15, opened 2018-01-08
RFC for GPS support in Cartographer.

spielawa:
Add RFC for GPS support in Cartographer. 

[Rendered](https://github.com/spielawa/rfcs/blob/gps_support/text/0000-gps_support.md)

PAIR=wohe


text/0000-gps_support.md
```diff
@@ -0,0 +1,24 @@
+# NavSatFix (GPS) Support in Cartographer ROS
+
+## Summary
+[summary]: #summary
+
+The goal is to add NavSatFix (GPS) support to Cartographer ROS.
```
SirVer: Add a link to http://docs.ros.org/jade/api/sensor_msgs/html/msg/NavSatFix.html

text/0000-gps_support.md
```diff
@@ -0,0 +1,24 @@
+# NavSatFix (GPS) Support in Cartographer ROS
+
+## Summary
+[summary]: #summary
+
+The goal is to add NavSatFix (GPS) support to Cartographer ROS. 
+The NavSatFix (GPS) data will be used in pose graph optimization, using the fixed frame pose feature of Cartographer.
```
SirVer: in *the* ?

text/0000-gps_support.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+In outdoor applications, especially cars, GPS data is commonly available. 
+Using this data will reduce the error in location, especially over large distances.
```
SirVer: s/error/drift/?

text/0000-gps_support.md
```diff
@@ -0,0 +1,24 @@
...
+
+In outdoor applications, especially cars, GPS data is commonly available. 
+Using this data will reduce the error in location, especially over large distances. 
+
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `nav_sat_sampling_ratio`.
+
+These messages will be translated into a common cartesian coordinate frame (e.g. ECEF) and given to Cartographer as `FixedFramePoseData`.
```
SirVer: I do not think e.g. is correct here, right? We will hardcode using ECEF for the moment. Could you add a paragraph on considered alternatives?

text/0000-gps_support.md
```diff
@@ -0,0 +1,24 @@
...
+## Motivation
+[motivation]: #motivation
+
+In outdoor applications, especially cars, GPS data is commonly available. 
+Using this data will reduce the error in location, especially over large distances. 
+
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `nav_sat_sampling_ratio`.
```
cschuet: Why do you want to feed the fixed-frame poses from a GPS specific message type? Maybe other people want to use a different message type. Aren't fixed-frame poses just time-stamped poses? Why not use a ROS msg which contains just that?

text/0000-gps_support.md
spielawa: done

text/0000-gps_support.md
spielawa: done

text/0000-gps_support.md
spielawa: done

text/0000-gps_support.md
wohe: `NavSatFix` is the message widely used in the ROS community, e.g. what a GPS driver provides. This is similar to the code we already have in Cartographer which understands `LaserScan` messages and translates them to the Cartographer-specific data format. `NavSatFix` is not GPS specific and supports other GNSS systems.

text/0000-gps_support.md
gaschler: Just to be clear, GPS support will not introduce any changes to cartographer, is unlikely to need any new optimization weights, just FixedFramePose is used as it is?

text/0000-gps_support.md
gaschler: Which additional dependencies will you need? I'm sure the conversion is in package we currently don't depend on.

cschuet: I see, but it is specific to satellite based positioning systems. I see the `NavSatStatus` submessage in there. Why not simply specify the fixed-frame pose as timed `geometry_msgs/Pose.msg`?

Most people use Cartographer for indoor robot navigation. In this setup it is very unlikely that your fixed-frame is coming from a satellite based system. Don't you think it is unfortunate that everyone now has to invent a `NavSatStatus` that is completely ignored  by your conversion code?

I am not opposed to using `NavSatFix` messages. I just want to make sure we  implement this feature in a way that it appeals to the general usecase.

wohe: I think it also makes sense to add a message that allows specifying fixed frame poses directly for other uses cases. To me this seems similar how Cartographer ROS already support 3 ways to provide range data.

This RFC is about adding GPS-like data. Providing a generic message instead of `NavSatFix`would be cumbersome (everyone would have to convert to a cartesian coordinate system) and as mentioned having a `/fix`topic is common in ROS.

Would you agree that supporting `NavSatFix` and a generic fixed frame pose message are two orthogonal improvements to Cartographer ROS?

cschuet: The drawback of this is that it requires the introduction of another `use_XXX` and `XXX_sampling_ratio`. Or I guess if you would use a more generic name than `use_nav_sat` we could use message introspection to handle both types of messages by the same option.

wohe: Shouldn't we keep this consistent with the way different `LaserScan`, `MultiEchoLaserScan` and `PointCloud2`. Also, the topic `/fix` is very specific to `NavSatFix`. We can rename the sampling ratio to `fixed_frame_pose_sampling_ratio` if you would like this to be forward looking to when a different way to handle fixed frame poses is added. Otherwise we would refactor it then.

cschuet: Sounds good, let's try to share the sampler then at least.

text/0000-gps_support.md
wohe: Of course we would like to improve Cartographer to make the most out of GPS, e.g. maintaining fixed frame pose optimization variables between optimizations. But this is not specific to the GPS support in any way.

text/0000-gps_support.md
wohe: This does not need any additional dependencies. Which conversion?

spielawa: We write the function for converting from GPS to cartesian coordinates ourselves.

gaschler: The conversion from latitude and longitude given in NavSatFix to a pose in FixedFrameData.
Will this be in cartographer_ros or in a library that we link (which one)?

cschuet (approved):

text/0000-gps_support.md
gaschler: @spielawa How complex will the conversion be, perhaps roughly in lines of code?

spielawa: The formulas are here: https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates, I've added them to the RFC.

text/0000-gps_support.md
```diff
@@ -0,0 +1,31 @@
...
+These messages will be [translated](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates) into a common cartesian coordinate frame, [ECEF](https://en.wikipedia.org/wiki/ECEF), and given to Cartographer as `FixedFramePoseData`.
+
+Alternatively, we could translate to a local ground reference frame ([ENU](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED)) with the z-axis pointing up, and the origin at the start position.
+
+## Discussion Points
+[discussion]: #discussion
+
+### Should we implement a generic fixed frame pose message instead?
+
+This RFC is specific to adding GPS-like data from statelite based positioning systems.
```
gaschler: satellite

gaschler (approved):

text/0000-gps_support.md
```diff
@@ -0,0 +1,31 @@
...
+Using this data will reduce the drift in location, especially over large distances. 
+
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `fixed_frame_pose_sampling_ratio`.
+
+These messages will be [translated](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates) into a common cartesian coordinate frame, [ECEF](https://en.wikipedia.org/wiki/ECEF), and given to Cartographer as `FixedFramePoseData`.
+
+Alternatively, we could translate to a local ground reference frame ([ENU](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED)) with the z-axis pointing up, and the origin at the start position.
```
SirVer: Could we settle on either ECEF or ENU? Maybe discuss the differences and advantages of both here, since both choices are rather opaque without more background.

spielawa:
PTAL

text/0000-gps_support.md
```diff
@@ -0,0 +1,36 @@
...
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `fixed_frame_pose_sampling_ratio`.
+
+These messages will be [translated](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates) first into a common cartesian coordinate frame, [ECEF](https://en.wikipedia.org/wiki/ECEF). 
+
+The we will translate to a local ground reference frame (e.g. [ENU](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED)) with the z-axis pointing up, and the origin at the start position.
+
+And the result will be given to Cartographer as `FixedFramePoseData`.
```
wohe: nit: Drop the 'And'?

text/0000-gps_support.md
```diff
@@ -0,0 +1,36 @@
...
+Using this data will reduce the drift in location, especially over large distances. 
+
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `fixed_frame_pose_sampling_ratio`.
+
+These messages will be [translated](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates) first into a common cartesian coordinate frame, [ECEF](https://en.wikipedia.org/wiki/ECEF). 
+
+The we will translate to a local ground reference frame (e.g. [ENU](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED)) with the z-axis pointing up, and the origin at the start position.
```
wohe: nit: Then we will transform the pose into a local...

text/0000-gps_support.md
```diff
@@ -0,0 +1,36 @@
...
+[discussion]: #discussion
+
+### Should we implement a generic fixed frame pose message instead?
+
+This RFC is specific to adding GPS-like data from statellite based positioning systems. 
+We use the `NavSatFix` message which is widely used in the ROS community, and convert it from longitude, latitude, altitude to a cartesian coordinate system.
+In addition, in a separate RFC, a generic fixed frame pose message could also be supported.
+
+### Why don't we use ECEF? 
+In cartographer, the z axis points up. We want to preserve that property also in the fixed frame pose.
```
wohe: typo: In Cartographer...

text/0000-gps_support.md
```diff
@@ -0,0 +1,36 @@
...
+Using this data will reduce the drift in location, especially over large distances. 
+
+## Approach
+[approach]: #approach
+
+We will wire up the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) and introduce two new options for this: `use_nav_sat` and `fixed_frame_pose_sampling_ratio`.
+
+These messages will be [translated](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates) first into a common cartesian coordinate frame, [ECEF](https://en.wikipedia.org/wiki/ECEF). 
+
+Then we will translate to a local ground reference frame (e.g. [ENU](https://en.wikipedia.org/wiki/Axes_conventions#Ground_reference_frames:_ENU_and_NED)) with the z-axis pointing up, and the origin at the start position.
```
wohe: "transform" instead of "translate" since this includes a change in orientation?

wohe (approved):

SirVer (approved):
lgtm

PR #16, opened 2018-01-09
Draft Collator Proposal.

cschuet:
[Rendered](https://github.com/cschuet/rfcs/blob/collator-interface/text/0000-collator-interface.md)

### DIFF
--- a/text/0008-collator-interface.md
+++ b/text/0008-collator-interface.md
@@ -1,0 +1,32 @@
+# Customizable Collators
+
+## Summary
+[summary]: #summary
+
+We propose to introduce a `sensor::CollatorInterface` and add two implementations of the same:
+
+1. one which sorts sensor data across sensor data across all trajectories (this is what the current implementation, `Collator`, does).
+2. one which sorts sensor data only within the trajectory they belong to, allowing other trajectories to make progress even if one is blocked.
+
+## Motivation
+[motivation]: #motivation
+
+For cloud-based mapping we have to deal with the situation where the robot to cloud uplink is unrealible.
+A robot might venture into an area where no Wi-Fi signal is available thereby severing the cloud uplink.
+The collator in the Cartographer cloud instance would currently stall as it is waiting for data from that robot before allowing any other trajectories to make progress.
+Ideally this behavior would be configurable through `MapBuilderOptions`.
+
+## Approach
+[approach]: #approach
+
+We propose to introduce a `sensor::CollatorInterface` which would be a pure virtualization of the current `Collator`.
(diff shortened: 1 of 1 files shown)


text/0000-collator-interface.md
```diff
@@ -0,0 +1,31 @@
...
+For cloud-based mapping we have to deal with the situation where the robot to cloud uplink is unrealible.
+A robot might venture into an area where no Wi-Fi signal is available thereby severing the cloud uplink.
+The collator in the Cartographer cloud instance would currently stall as it is waiting for data from that robot before allowing any other trajectories to make progress.
+Ideally this behavior would be configurable through `MapBuilderOptions`.
+
+## Approach
+[approach]: #approach
+
+We propose to introduce a `sensor::CollatorInterface` which would be a pure virtualization of the current `Collator`.
+The current `Collator` would be renamed to `AcrossTrajectoryCollator` and be marked as an implementation of that interface.
```
cschuet: If anyone has a better names for the implementations I'd be grateful.

text/0000-collator-interface.md
```diff
@@ -0,0 +1,31 @@
...
+A new `Collator` would be introduce, `PerTrajectoryCollator`, which only collates sensor data belonging to the same trajectory.
```
pifon2a: nit: "introduced"

text/0000-collator-interface.md
pifon2a: Will we need 'AcrossTrajectoryCollator' at all? Is it possible to just replace the current Collator with 'PerTrajectoryCollator'? Or implement the approach you suggested and then compare results and drop 'AcrossTrajectoryCollator' if we want it to be done safely.

pifon2a (approved):

SirVer:
@cschuet The rendered link should contain the name of the branch, not a particular SHA so that it always is current to the latest in this PR. I fixed this for you here.

SirVer:
lgtm in general. Adding @wohe, maybe he has feelings about this.

SirVer (approved):

wohe:
I think the RFC should mention that per-trajectory collating means that there no longer is a deterministic result given the same input data. Or do I misunderstand something here?

cschuet:
@wohe I think you are right about that.
@pifon2a I think this is also the answer to your question: we want to retain the old trajectory builder to keep the possibility of having a deterministic result.

gaschler:
LGTM. I also agree with wohe that the non-determinism of PerTrajectoryCollator should be documented.

gaschler (approved):

PR #17, opened 2018-01-11
AMCL replacement RFC

doisyg:
[AMCL replacement RFC](https://github.com/doisyg/rfcs/blob/master/text/0000-amcl-replacement.md)


ojura:
Please put a [rendered](https://github.com/doisyg/rfcs/blob/master/text/0000-amcl-replacement.md) link in the PR description:
`[Rendered](https://github.com/doisyg/rfcs/blob/master/text/0000-amcl-replacement.md)`

MichaelGrupp:
Please also change to stick to this guideline (see README):
> Make sure to have one sentence per line in the RFC as it makes reading diffs much easier. There is no characters per line limit for RFCs.

text/0000-amcl-replacement.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Summary
+[summary]: #summary
+
+Make cartographer in Pure localization mode works as a replacement for AMCL and be one of the ROS localization method.
+
+## Motivation
+[motivation]: #motivation
+
+Cartographer is really powerful in term of localization and a pure localization mode was developed (it works already great). In order to be adopted by the ROS community it should be simple to run and works similarly  in its implementation as AMCL
```
SirVer, line 11: nit: two spaces after similarly

text/0000-amcl-replacement.md
```diff
@@ -0,0 +1,41 @@
...
+
+* Support the `/initialpose` topic to relocate the robot. 
+This topic is for instance used out of the box by rviz `2D pose estimate` button (or keyboard shortcut: p). 
+A quick and dirty implementation was done here: https://github.com/googlecartographer/cartographer_ros/pull/637
+
+
+* Publish the static map against which localization is done on the `/map` topic for the navigation stack to be able to be run on top of it (global planner for instance) and to have a static reference to record waypoint in.
+Currently with the occupancy_grid node, the map which is published is a superposition of the static map + the current submap. 
+A workaround is to not run the occupancy_grid node and in parallel run a `map_server` node which publish a *.pgm/*.png/ + *.yaml map that had to be generated from the same *.pbstream file which used by the pure localization node. 
+It would be nice to have everything work in one step.
```
SirVer, line 28: I agree that this should be done in one step. I suggest adding options to blacklist or whitelist trajectories from the occupancy_grid_node, so that only the map is published and not the pure localization trajectories.

text/0000-amcl-replacement.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+* Provide a way to easily edit a map afterwards (if this is possible).
```
SirVer, line 31: This is difficult for Cartographer submaps. I think it would require a separate RFC with more design.

As a workaround, a map can be saved into pgm yaml, modified in gimp and then served using `map_server`.

text/0000-amcl-replacement.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+* Publish the static map against which localization is done on the `/map` topic for the navigation stack to be able to be run on top of it (global planner for instance) and to have a static reference to record waypoint in.
+Currently with the occupancy_grid node, the map which is published is a superposition of the static map + the current submap. 
+A workaround is to not run the occupancy_grid node and in parallel run a `map_server` node which publish a *.pgm/*.png/ + *.yaml map that had to be generated from the same *.pbstream file which used by the pure localization node. 
+It would be nice to have everything work in one step.
+
+
+* Provide a way to easily edit a map afterwards (if this is possible). 
+AMCL maps being bitmaps, there are easily cleaned manually, in GIMP for instance.
```
SirVer, line 32: nit: the term should be ROS map, since it is a ros concept, not an AMCL concept.

text/0000-amcl-replacement.md
```diff
@@ -0,0 +1,41 @@
...
+A workaround is to not run the occupancy_grid node and in parallel run a `map_server` node which publish a *.pgm/*.png/ + *.yaml map that had to be generated from the same *.pbstream file which used by the pure localization node. 
+It would be nice to have everything work in one step.
+
+
+* Provide a way to easily edit a map afterwards (if this is possible). 
+AMCL maps being bitmaps, there are easily cleaned manually, in GIMP for instance. 
+This is very useful when maping a site which has moveable furnitures (chairs, people standing, christmas tree, etc...), we want to be able to remove these non fix objects from the localization map.
+
+
+* Make the origin of the map changeable.
```
SirVer, line 36: Not sure why this is in here, there is no motivation for this provided. What does the origin mean even?

SirVer:
I agree that this a reasonable step to integrate better with ROS, so I think the RFC is a great contribution. I feel it is very clear and strong in its initial points and fizzles out towards the end - I left comments accordingly.

text/0000-amcl-replacement.md
doisyg: I feared that it was not easily doable. Modifying a pgm yaml and exposing it with map_server would only be okay for use by the planner, but these changes won't be used by the cartographer localizer.

text/0000-amcl-replacement.md
doisyg: Yes, but I wanted to emphasis that ROS bitmap maps is the format used by AMCL. I will rephrase.

text/0000-amcl-replacement.md
doisyg: Sorry for being unclear. Currently with cartographer in localization mode, the pose of the robot is defined in the map frame relative to a point with the (0 0) coordinate. This point being located on the map to where the robot started the maping procedure (ie where the robot was when the bag used to generate the .pbstream started). This is not very conveniant as every other point of interest have to be defined relative to this origin, and this origin will change if a new map is built. A workaround is to publish a tf from the origin of the cartographer map to a choosen physical point and then to record any point of interest relative to this physical point. But I think this could be simplified.

doisyg:
Thanks for the feedback. I replied to your comments. How should we proceed with the discussion on this RFC? Here is a good place?

text/0000-amcl-replacement.md
cschuet: Adding a blacklist sounds good to me. @doisyg Can you update the RFC and officially make this the plan? (if you agree that this solves your usecase)

text/0000-amcl-replacement.md
cschuet: I would also remove this from this RFC. This is a bigger endeavor.

text/0000-amcl-replacement.md
cschuet: @doisyg Can you make a concrete proposal what you want to modify in the code?

dan9thsense:
Is this still alive?  It would be an outstanding feature.

text/0000-amcl-replacement.md
cschuet: @sebastianklose This could be another feature for the pbstream tool we discussed.

text/0000-amcl-replacement.md
MichaelGrupp: Update: this "black/whitelist" feature was merged in https://github.com/googlecartographer/cartographer_ros/pull/899 (--> https://github.com/googlecartographer/cartographer_ros/commit/f5b583fde8d857f72c45da3820389b4227d3b084)

text/0000-amcl-replacement.md
MichaelGrupp: How should this work? 

1. Start mapping / localizing at an initial pose that's relative to a world-fixed parent frame - for me this is not a workaround as you say, but the only way to do it right. Your motivation sounds like you have such a global reference frame anyway, otherwise it wouldn't matter?

2. You can't simply change the trajectory afterwards to start at an arbitrary point without adding a proper parent frame - either you move the whole trajectory inside the map (semantically wrong) or keep it as it is and add an additional starting node (adds a jump that can be problematic for optimization). Or do I get it wrong?

text/0000-amcl-replacement.md
MichaelGrupp: And: another, very efficient alternative to publish a static map is the `pbstream_map_publisher` which I added in the meantime ([code](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/pbstream_map_publisher_main.cc)). Just run:
```
rosrun cartographer_ros cartographer_pbstream_map_publisher --pbstream_filename xyz.pbstream --resolution 0.x --map_frame_id foo --map_topic bar
```

MichaelGrupp:
My cents: 

IMO, most of this could be "solved" by improving the documentation of `cartographer_ros` e.g. by adding guidelines or tutorials how to use it with off the shelf ROS tools. Because the building blocks for the first three points are already there. I agree though that it's absolutely not obvious for people who just want to _use_ Cartographer at the moment.

Initial pose:
  * start pose is available in `trajectory_builder_options`
  * services like `/get_trajectory_states`, `/finish_trajectory` and `/start_trajectory` are already there and can be used to implement an `/initialpose` handler

Convenient map publishing:
  * `pbstream_map_publisher` executable
  * `occupancy_grid_node` options to select by frozen / unfrozen state

text/0000-amcl-replacement.md
doisyg:
```
1.    Start mapping / localizing at an initial pose that's relative to a world-fixed parent frame - for me this is not a workaround as you say, but the only way to do it right. Your motivation sounds like you have such a global reference frame anyway, otherwise it wouldn't matter?
```
I am not sure I understand fully your comment. But we do not necessarily have a global reference frame. I know a couple of use cases (including ours) where a map is build first, then coordinates of point of interests are recorded/defined . Is that case, the ROS map is the global reference frame (or its origin). The advantage of the ROS map implementation is that by default, the origin of the map is the corner of the map, and that is it is very easy to change where this 0 is located with the companion yaml file. http://wiki.ros.org/map_server#YAML_format

```
2.    You can't simply change the trajectory afterwards to start at an arbitrary point without adding a proper parent frame - either you move the whole trajectory inside the map (semantically wrong) or keep it as it is and add an additional starting node (adds a jump that can be problematic for optimization). Or do I get it wrong?

```
As I said before, a ROS user could add a tf transform on top of cartographer to get a convenient 0 on its 'map_relocated' frame, but I feel that there should be an easier way.
For instance, I can imagine 3 parameters (x,y, theta) that the cartographer_ros node will use to offset the 0 of the map when publishing the robot pose in ROS.

doisyg:
I actually made a quick and dirty /initialpose handler few month ago in this PR: https://github.com/googlecartographer/cartographer_ros/pull/637
This is what pushed me to write the RFC.
I didn't have time to properly take care of it recently so I don't know it is still working with the current version of cartographer. I may have a look back at it this summer.

text/0000-amcl-replacement.md
MichaelGrupp: What I mean is that you always need an additional frame in which your SLAM trajectories are embedded for your use cases.

> a ROS user could add a tf transform on top of cartographer to get a convenient 0 on its 'map_relocated' frame, but I feel that there should be an easier way.
For instance, I can imagine 3 parameters (x,y, theta) that the cartographer_ros node will use to offset the 0 of the map when publishing the robot pose in ROS

The only difference would be that the additional TF is provided by the `cartographer_node`, which is a kind of special thing to add to its options. Just add a `static_transform_publisher` to your launch file - done.

NTUwilliam:
Hi all!

Firstly, thank you all again for your great works ! Cartographer is the best 2D SLAM algorithm I've ever used. 
I browsed through this RFC & other related issues and found that I may have the same problem. 

My issue is like this:
I've been using 2D Cartographer **pure localization** for a couple of months and I found that in some situations that it would **take too much time** (more than 40 seconds) for Cartographer to achieve the correct initial localization. Sometimes even **can not** achieve correct initial localization in the large symmetric environment, which is featureless and is indeed a challenge for every SLAM / Localization algorithm. 

The _**failed** localization_ process is shown as .gif file below:
fail


Part of the _**40 seconds** initial localization_ process is shown as .gif file below:
long-time-success


So I'm thinking about providing the **initial pose** for Cartographer to speed up and get the correct **initial pure localization**. The initial pose can be provided by other **coarse localization methods**, like Magnetic Field / WiFi localization. So I'm wondering:

1. Which part of the code is used for the **initialization of pure localization**?

2. How can I modify the `trajectory_builder_options` to incorporate my own initial pose but doesn't influence the normal laser scan matching localization in the later phase? 

   I've browsed through all the codes and found that the part which maybe related is the `ComputeLocalToGlobalTransform` part in [pose_graph_2d.cc](https://github.com/googlecartographer/cartographer/blob/2bd987ffb40e2d6667682c5c96432173f371094f/cartographer/mapping/internal/2d/pose_graph_2d.cc#L896) ? 
And I'm wondering what's the meaning of the [data_](https://github.com/googlecartographer/cartographer/blob/2bd987ffb40e2d6667682c5c96432173f371094f/cartographer/mapping/internal/2d/pose_graph_2d.cc#L902) ? I only found something related in [ceres_pose.h](https://github.com/googlecartographer/cartographer/blob/2bd987ffb40e2d6667682c5c96432173f371094f/cartographer/mapping/internal/optimization/ceres_pose.h#L47) but I don't think it's the right explanation for the `data_` that I'm looking for.

   If not this `ComputeLocalToGlobalTransform` part, which file should I modify to input the known initial pose for pure localization?


3. I suppose `trajectory_id = 0` means the robot trajectory in the mapping phase (in the .pbstream file) and the `trajectory_id = 1` means the robot trajectory in the real-time localization phase, is that right? 


Any hint or advice is appreciated. Thanks a lot !!
@MichaelGrupp @SirVer @ojura @cschuet @doisyg

ojura:
You can use the initial pose option with the `start_trajectory` node. Fire up an online node with prior map __without__ starting a default trajectory (`-start_trajectory_with_default_topics=false`), and then use the `start_trajectory` node (which accepts an initial pose).

WilliamWoo45:
@ojura Thanks for your speedy reply.

The main file for `start_trajectory` node is this [start_trajectory_main.cc](https://github.com/googlecartographer/cartographer_ros/blob/29105294468e0cc81def94fd68a540eef7533f46/cartographer_ros/cartographer_ros/start_trajectory_main.cc) , right?


I modified the `start_trajectory` node like this [start_trajectory_main_modified.cc](https://github.com/NTUwilliam/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/start_trajectory_main.cc#L41) (add the initial translation and rotation info). 
Then I changed the `start_trajectory_with_default_topics` to **false** in [node_main.cc](https://github.com/NTUwilliam/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/node_main.cc#L41) and [node_grpc_main.cc](https://github.com/NTUwilliam/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/cartographer_grpc/node_grpc_main.cc). These two files are the only files that I can find related to 
`start_trajectory_with_default_topics` 

After all these modifications, I run pure localization and find that the initial pose is still not correctly setted (exactly same result as before).


> Fire up an online node with prior map

And I don't quite understand this. Could you elaborate on how to achieve this?
Am I missing something else?

Thanks!

ojura:
You should not have to change any code. These are Google flags, and you specify them on the command line of the executable (`-start_trajectory_with_default_topics=true` with the online node, for example).

For pure localization, you need to specify a prior map .pbstream file (specified with `load_state_filename` flag with the online node). You should also specify `start_trajectory_with_default_topics=false`. And then run the `start_trajectory` with the `initial_pose` option. See [this](https://github.com/googlecartographer/cartographer_ros/issues/579#issuecomment-351544734) post for an example.

NTUwilliam:
@ojura Thanks!!!

I'll try it asap.

WilliamWoo45:
@ojura 

> You should also specify `start_trajectory_with_default_topics=false`.

1. Do you mean add the code `start_trajectory_with_default_topics=false` into the demo_backpack_2d_localization.launch file like [this](https://github.com/NTUwilliam/cartographer_ros/blob/master/cartographer_ros/launch/demo_backpack_2d_localization.launch#L31) ?

> See this post for an example.

2. I've tried to implement @doisyg 's method mentioned by you in [this](https://github.com/googlecartographer/cartographer_ros/issues/579#issuecomment-352537445) post but it's not working :(

> For pure localization, you need to specify a prior map .pbstream file (specified with load_state_filename flag with the online node). You should also specify start_trajectory_with_default_topics=false. And then run the start_trajectory with the initial_pose option.

3. I also tried your idea above. Firstly launch **pure localization** according to the example steps in the [Cartographer Docs](https://google-cartographer-ros.readthedocs.io/en/latest/demos.html) then do `rosrun cartographer_ros cartographer_start_trajectory ... -initial_pose '{to_trajectory_id = 0, relative_pose = { translation = { 0., 0., 0. }, rotation = { 0., 0., 0., } } }`. But it **failed** to set the initial pose for the reason shown below: 
```
williamwoo@williamwoo:~$ rosrun cartographer_ros cartographer_start_trajectory ... -initial_pose '{to_trajectory_id = 0, relative_pose = { translation = { 20, 0, 0}, rotation = {0., 0., 0., } } }'
F0303 16:08:11.567324 30785 start_trajectory_main.cc:115] Check failed: !FLAGS_configuration_directory.empty() -configuration_directory is missing.
*** Check failure stack trace: ***
    @     0x7f0e7594a5cd  google::LogMessage::Fail()
    @     0x7f0e7594c433  google::LogMessage::SendToLog()
    @     0x7f0e7594a15b  google::LogMessage::Flush()
    @     0x7f0e7594ce1e  google::LogMessageFatal::~LogMessageFatal()
    @           0x522142  main
    @     0x7f0e728f7830  __libc_start_main
    @           0x523b09  _start
    @              (nil)  (unknown)
Aborted (core dumped)
```

So now I'm really confused about these different **steps** for setting `initial_pose` for **pure localization mode**. Could you elaborate on the **correct steps**?

Sincerely thanks!!

ojura:
Regarding 1, yes, but do not put spaces around =.

Regarding the last error message, it is clear what is missing. Please carefully examine the error messages. You did not specify the configuration directory.

WilliamWoo45:
@ojura 
Thanks for all your guidance. I've made it work!!

I think the correct **steps** for setting `initial_pose` for **pure localization mode** should be like this:

1. Add the code `start_trajectory_with_default_topics=false` into the demo_backpack_2d_localization.launch file like [this](https://github.com/NTUwilliam/cartographer_ros/blob/master/cartographer_ros/launch/demo_backpack_2d_localization.launch#L31)


2. Start cartographer node in pure localization mode by:
```
roslaunch cartographer_ros demo_backpack_2d_localization.launch \
   load_state_filename:=${HOME}/xxx.bag.pbstream \
   bag_filename:=${HOME}/xxx.bag
```

2. Set the initial pose in **another Terminal** right after step 1:
`rosrun cartographer_ros cartographer_start_trajectory ... -configuration_directory '${HOME}/xxx/xxx/cartographer_ros/cartographer_ros/configuration_files' -configuration_basename 'backpack_2d_localization.lua' -initial_pose '{to_trajectory_id = 0, relative_pose = { translation = { 0, 0, 0}, rotation = {0., 0., 0., } } }'`

MarcelSoler:
Hello there! I'm not sure if this is the correct place to ask this but because if aligned with the discussion above I'll post it here.

When I use start_trajectory with a pose the robot coordinates (in map frame) doesn't match where the robot is placed. In this issue there are more details:

https://github.com/googlecartographer/cartographer/issues/1477

What I would like to know if there some kind of parameter to ensure map and trajectory 0 are in the same system of coordinates after being optimised.  

Thank you!

EDIT:
Solved  adding "timestamp = 0".

PR #18, opened 2018-01-11
Add an RFC for landmark support.

pifon2a:
[rendered version](https://github.com/pifon2a/rfcs/blob/master/text/0011-landmarks.md)


text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+Landmarks translation/orientation can improve SLAM quality. Landmarks rendered on the map can be useful for inspection. 
+
+## Approach
+[approach]: #approach
+Landmarks are sparse features of geometry that we want to map. Unlike dense range data, landmarks are assumed to be sparse in comparison. In order to support using them, current SLAM approach has to be generalized. 
+
+### Sensor
+Each landmark will have an ID and a position relative to the robot with or without orientation. A landmark observation can contain several landmarks at a time.
+
+### Submap
```
cschuet: You should make the connection how you want to go from timed landmark observations relative to the robot to adding them to the submap.

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+Submaps for 2D and 3D geometries are occupancy grids that contain accumulated range data.  There can be two approaches to add landmarks to the submaps:
+
+1. Submap holds a dictionary with landmark data keyed by landmark IDs;
```
cschuet: I assume when you talk about landmark data this includes the landmark pose within the submap frame? If so you should make that more explicit.

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+2. Every cell of the grid is modified to hold various features, for example, likelihood of landmarks, ML features.
```
cschuet: So to rephrase that: 1. is something like a maximum posterior approach, while with 2. you want to model the distribution. How would you use this distribution knowledge to approve mapping? How would e.g. the ContraintBuilder and PoseGraph use this information?

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+### Constraint builder
+`ConstraintBuilder` should use landmarks while adding intra-submap constraints:
+
+* `FastCorrelativeScanMatcher` will use landmarks in the bounding function;
+* `CeresScanMatcher` will have an additional term in its cost function.
+
+### Optimization problem
+Landmarks may enter sparse pose graph optimization as additional independent variables. It would increase the dimension of the optimization by `number_of_landmarks * number_of_parameters_per_landmark`, which could be undesirable when there is a significant number of landmarks.
+
+There is another use case here. If Cartographer is provided with the relative positions of landmarks with respect to each other, it can be added to the sparse pose graph optimization similarly to `FixedFramePoseData`.
```
cschuet: I would say that knowing the relative position of two landmarks should be added as another constraint between the location variables of the two landmarks.

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Submap
+Submaps for 2D and 3D geometries are occupancy grids that contain accumulated range data.  There can be two approaches to add landmarks to the submaps:
+
+1. Submap holds a dictionary with landmark data keyed by landmark IDs;
+2. Every cell of the grid is modified to hold various features, for example, likelihood of landmarks, ML features. 
+
+The second approach is more generic and makes it possible to introduce landmarks based on the strength of the signal, e.g. Wi-Fi signal, BT beacons. The first approach should work faster, it is also more flexible and non-intrusive.
+  
+### Constraint builder
```
cschuet: In general I think you should be more concrete what changes you are planning to make to `ConstraintBuilder` and `PoseGraph`.

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+`ConstraintBuilder` should use landmarks while adding intra-submap constraints:
+
+* `FastCorrelativeScanMatcher` will use landmarks in the bounding function;
+* `CeresScanMatcher` will have an additional term in its cost function.
+
+### Optimization problem
```
cschuet: IIUC you are seeing the addition of landmarks to the optimization problem as optional? I think this RFC would be more readable if you outlined your favorite approach in greater detail and then in a separate section explain what alternatives are possible.

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
+# Landmark observations
+Add landmarks support to Cartographer.
+
+## Summary
+[summary]: #summary
+Cartographer will be able to work with a new type of sensor data that contains IDs and locations of landmarks.
+
+## Motivation
+[motivation]: #motivation
+Landmarks translation/orientation can improve SLAM quality. Landmarks rendered on the map can be useful for inspection.
```
MichaelGrupp: I'd change to: "Landmark visualizations can be useful for inspection."

Rendered on the map sounds like the occupancy grid would change.

text/0000-landmarks.md
MichaelGrupp: What is an  "ML feature" in this context?

text/0000-landmarks.md
```diff
@@ -0,0 +1,43 @@
...
+Landmarks may enter sparse pose graph optimization as additional independent variables. It would increase the dimension of the optimization by `number_of_landmarks * number_of_parameters_per_landmark`, which could be undesirable when there is a significant number of landmarks.
+
+There is another use case here. If Cartographer is provided with the relative positions of landmarks with respect to each other, it can be added to the sparse pose graph optimization similarly to `FixedFramePoseData`.
+
+### Rendering
+Landmarks IDs and positions should be propagated to RViz for rendering.
+
+## Discussion Points
+[discussion]: #discussion
+Should the landmarks be a part of local SLAM if a robot is capable to detect landmarks while moving?
```
MichaelGrupp: I can imagine landmark observations to introduce significant jumps that we want to avoid in local SLAM. Think of the case where the local SLAM front-end is used as odometry.

I see the incorporation of landmarks more as a corrective measure to be used in the backend, similar to the way loop closures are utilized.

MichaelGrupp:
Added a few comments.

text/0000-landmarks.md
pifon2a: Agreed, especially for the Magazino use case. It would be interesting to experiment with it later though.

text/0000-landmarks.md
pifon2a: To Michael: any ML features, for example, if robot can recognize bounding boxes of different objects while building a map, etc. Or features for visual SLAM.

pifon2a: To Christoph: the approach is the same, it is more about how we store landmarks in Submap class: do we store them separately from the probability grid or not.

text/0000-landmarks.md
MichaelGrupp: Sure, but it shouldn't happen implicitly.

text/0000-landmarks.md
cschuet: Oh are you saying that in the dictionary approach you would store for a given landmark different possible poses along with a probability also?

text/0000-landmarks.md
cschuet: We know that scan matching already introduces such jumps, which is why the odom topic Cartographer publishes is at times non-smooth and not great for feeding into navstacks.

@MichaelGrupp Are you saying that you expect that landmarks will make those existing jumps bigger?

MichaelGrupp: Ok maybe let's forget the odom use case. I mainly struggle with how landmarks that provide global information can be used in local SLAM (which to my understanding should estimate the local motion of the robot).

Let's think of the hypothetical situation where the system observes a landmark that can be associated with an absolute pose inside the global map with high certainty (e.g. some accurately measured marker). While this is certainly beneficial for the optimization to produce a globally consistent map, how should this be incorporated in local SLAM? Imagine we have collected quite a large amount of drift at that point and observe the landmark that tells us we're actually off by a few meters.

pifon2a: In local SLAM it would be used exactly like we are using range data now: we get new a new point cloud and optimize translation/rotation w.r.t. the probability grid (2d) / hybrid grid (3d). Assume we have landmark(s) that are observed by robot in every trajectory node inside one of the submaps. Then we would try to maximize f_1(range_data (translation, rotation), grid) + f_2(landmarks_visible_from_trajectory_node, accumulated_landmarks_in_submap), i.e. it would require another criterion in the goal function.

It is definitely not what the use case you want. This would be a more generic approach, but it seems that it is not needed yet. It can be added later if needed. For now, I am removing local SLAM parts from the RFC. I'll expand the global SLAM part tomorrow.

text/0000-landmarks.md
pifon2a: Yes, constraints between pose graph nodes.

text/0000-landmarks.md
```diff
@@ -0,0 +1,29 @@
...
+## Motivation
+[motivation]: #motivation
+Landmarks translation/orientation can improve SLAM quality. Landmark visualizations can be useful for inspection.
+
+## Approach
+[approach]: #approach
+Landmarks are sparse features of geometry that we want to map. Unlike dense range data, landmarks are assumed to be sparse in comparison. In order to support using them, current SLAM approach has to be generalized. 
+
+### Sensor
+Each landmark will have an ID and a position relative to the robot with or without orientation. A landmark observation can contain several landmarks at a time.
```
wohe: It is also important to allow "observations" with zero landmarks seen, so that SLAM can continue, right?

text/0000-landmarks.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Optimization problem
+Landmarks may enter sparse pose graph optimization as additional independent variables. It would increase the dimension of the optimization by `number_of_landmarks * number_of_parameters_per_landmark`, which could be undesirable when there is a significant number of landmarks.
```
wohe: Isn't it the most important part that landmarks are in the pose graph? Why say "may" then?

text/0000-landmarks.md
wohe: This is different from `FixedFramePoseData` then? Observations are seen from the robot, not external observations from the robot, so `FixedFramePoseData` cannot be used as the error terms we have to compute are different, or do I misunderstand?

text/0000-landmarks.md
MichaelGrupp: Ahh ok, we are probably talking about different things. For me, the first thing that came to my mind was: landmark == semantic point in the environment with known location. The term landmark is a bit overused in SLAM :stuck_out_tongue: 

But looking again at your draft:
> Landmarks are sparse features of geometry that we want to map

So you mean a landmark is a local observation that can be tracked (similar to e.g. corner features in feature-based visual SLAM)? Then why not call it features if landmark is just an alias for "sparse features of geometry" to avoid confusion?

text/0000-landmarks.md
pifon2a: Yes, that's why we are sending a vector of landmarks that can be empty.

wohe: Could you make this a bit clearer in the RFC? How about
```
Each landmark observation contains all landmarks for which a relative position is currently observed. This can be an empty list.
```
?

text/0000-landmarks.md
pifon2a: Done

text/0000-landmarks.md
pifon2a: Sorry for the confusion. Of course the error terms are different. I thought that the principle in all residual blocks in sparse pose optimization was very similar: we compute relative poses and penalize the difference.

text/0000-landmarks.md
pifon2a: Done

text/0000-landmarks.md
pifon2a: Done

text/0000-landmarks.md
pifon2a: Thanks, Michael. When I first started writing this, I had an idea of covering all possible landmark use cases(semantic points and sparse features), but then it looks like it is unnecessary at this point. I will concentrate on "semantic points" for now.

MichaelGrupp: Ok, thanks for the clarification. Makes also more sense in my opinion. Looking forward to see the final RFC - this definitely something I'd be also interested in.

MichaelGrupp:
By the way: could you please add the link to the rendered version in the first comment? Makes it faster to see the newest full version when a notification comes in.
```
[rendered version](https://github.com/pifon2a/rfcs/blob/master/text/0000-landmarks.md)
```

pifon2a:
To Wolfgang: Could you please take another look?

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+[approach]: #approach
+Landmarks are sparse features of geometry that we want to map. Unlike dense range data, landmarks are assumed to be sparse in comparison. In order to support using them, current SLAM approach has to be generalized. 
+
+### Sensor
+Each landmark observation contains all landmarks for which a relative position is currently observed. This can be an empty list.
+
+```
+struct Landmark {
+  std::string id;
+  transform::Rigid3d transform;
```
wohe: I would like to see the code move towards being more explicit about the meaning of transforms, see https://google-cartographer.readthedocs.io/en/latest/terminology.html. How about `transform::Rigid3d landmark_to_tracking_transform`?

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+  LandmarkObservation {
+    common::Time time;
+    int trajectory_id;
+    
+    transform::Rigid3d transform;
+    double translation_weight;
+    double rotation_weight;
+  };
+  std::vector<LandmarkObservation> landmark_observations;
+  transform::Rigid3d global_pose;
```
wohe: nit: `global_landmark_pose`?

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+### Optimization
+Landmarks will enter sparse pose graph optimization as additional independent variables. It increases the dimension of the optimization by `number_of_landmarks * number_of_parameters_per_landmark`, which could be undesirable when there is a significant number of landmarks. 
+
+#### Cost function
+Landmark sensor data contains only the landmark position relative to the robot. In order to formulate a landmark-to-trajectory cost functional the global pose of the robot at the moment of the observation has to be known. Since a landmark observation can be made at the time between two trajectory nodes and, moreover, the global poses of the trajectory nodes are state variables as well, the translation and rotation of the landmark at the moment can be interpolated using the previous and the next node, i.e. taking a convex combination of the translation components and slerp of the quaternions. This logic is already implemented within `TransformInterpolationBuffer`.
```
wohe: It is correct that interpolation between trajectory nodes has to be done. This is not implemented yet as far as I know, though. Since trajectory node poses are optimization variables, interpolation must be implemented in a way that allows for automatic differentiation.

Also: Break line after each sentence?

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Cost function
+Landmark sensor data contains only the landmark position relative to the robot. In order to formulate a landmark-to-trajectory cost functional the global pose of the robot at the moment of the observation has to be known. Since a landmark observation can be made at the time between two trajectory nodes and, moreover, the global poses of the trajectory nodes are state variables as well, the translation and rotation of the landmark at the moment can be interpolated using the previous and the next node, i.e. taking a convex combination of the translation components and slerp of the quaternions. This logic is already implemented within `TransformInterpolationBuffer`. 
+
+Residuals depend on the parameter block `[c_previous, c_next, c_landmark]`, i.e. on the global poses of the previous and next trajectory nodes and the global pose of the landmark itself:
+
+![component diagram](0000-assets/cost.png "Cost function")
```
wohe: In principle correct. The computation of the error should be done in the tracking frame, and the "difference" to the expected pose must be done using by applying the appropriate inverse transform.

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+
+where `w` is the weight vector; `c` denotes global poses and `h` relative ones. 
+
+There will be no loss function as for intra submap constraints.
```
wohe: I guess this is to keep the implementation simple for now? It seems conceivable that a sensor might generate false landmark observations.

text/0000-landmarks.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Constraints
+The cost function for landmark-to-trajectory constraints depends on the previous and next trajectory nodes. They are not known at the moment when the sensor data comes, but we can find them prior to the `RunOptimization` call using `AddWorkItem()` for landmark observations that were not yet processed.
+
+There are two ways of storing the landmark constraints in `PoseGraph`:
+
+1.  as a separate `std::vector<LandmarkConstraint> landmark_constraints_`, which would bloat the `PoseGraph` even more. Since the next RFC will talk about adding landmark-to-landmark constraints, I would try to avoid this option.
+1. in the current `std::vector<Constraint> constraints_`, which would lead to expanding the `Constraint` struct and adding  a new value to `enum Tag { INTRA_SUBMAP, INTER_SUBMAP } tag;`
+
+I think the second option is a bit cleaner. Both are ugly though.
```
wohe: Can you explain why this is needed? It seems the map of `LandmarkNodes` which contains the observation already has all the necessary data. Why duplicate?

text/0000-landmarks.md
pifon2a: Sure, let's keep it simple and then adjust.

text/0000-landmarks.md
pifon2a: It contains all necessary data except IDs of the previous and next nodes. Another solution would be to add these IDs to LandmarkNode proto and then, when a Landmark is observed, add it to some queue of landmarks for which we do not know previous/next nodes yet.

jihoonl:
Do you plan to provide a separate topic to publish optimized land mark pose in global frame? It would be useful to use for navigation or other purposes.

text/0000-landmarks.md
```diff
@@ -0,0 +1,99 @@
...
+Landmarks are sparse features of geometry that we want to map.
+Unlike dense range data, landmarks are assumed to be sparse in comparison.
+In order to support using them, current SLAM approach has to be generalized.
+
+### Sensor
+Each landmark observation contains all landmarks for which a relative position is currently observed.
+This can be an empty list.
+
+```
+struct Landmark {
```
cschuet: I would rename this to `LandmarkObservation` to make the difference to `LandmarkNode` more explicit.

text/0000-landmarks.md
```diff
@@ -0,0 +1,99 @@
...
+
+![component diagram](0000-assets/cost.png "Cost function")
+
+where `w` is the weight vector; `c` denotes global poses and `h` relative ones.
+
+There will be no loss function as for intra submap constraints.
+
+#### Constraints
+The cost function for landmark-to-trajectory constraints depends on the previous and next trajectory nodes.
+They are not known at the moment when the sensor data comes, but we can find them prior to the `RunOptimization` call using `AddWorkItem()` for landmark observations that were not yet processed.
```
cschuet: @wohe Is it really necessary to do the MapById::upper_bound() call in a work item? Why not do them synchronously (for those new LandmarkNodes that we haven't previously determined the previous and next TrajectoryNode yet) in RunOptimization?

pifon2a:
@jihoon: sure.

cschuet (approved):

text/0011-landmarks.md
```diff
@@ -0,0 +1,99 @@
...
+
+![component diagram](0011-assets/cost.png "Cost function")
+
+where `w` is the weight vector; `c` denotes global poses and `h` relative ones.
+
+There will be no loss function as for intra submap constraints.
+
+#### Constraints
+The cost function for landmark-to-trajectory constraints depends on the previous and next trajectory nodes.
+They are not known at the moment when the sensor data comes, but we can find them prior to the `RunOptimization` call using `AddWorkItem()` for landmark observations that were not yet processed.
```
wohe: @cschuet Yes, this cannot be done outside `AddWorkItem()` for thread safety and lock contention. We do not have to do it prior to `RunOptimization()` though. This is the part I do not understand: Why not compute the previous and next trajectory node in the optimization problem and add the appropriate cost function? What would be the benefit of precomputing `LandmarkConstraint`s that warrants this additional complexity?

text/0011-landmarks.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+### Optimization
+Landmarks will enter sparse pose graph optimization as additional independent variables.
+It increases the dimension of the optimization by `number_of_landmarks * number_of_parameters_per_landmark`, which could be undesirable when there is a significant number of landmarks.
+
+#### Cost function
+Landmark sensor data contains only the landmark position relative to the robot. In order to formulate a landmark-to-trajectory cost functional the global pose of the robot at the moment of the observation has to be known.
+Since a landmark observation can be made at the time between two trajectory nodes and, moreover, the global poses of the trajectory nodes are state variables as well, the translation and rotation of the landmark at the moment can be interpolated using the previous and the next node, i.e. taking a convex combination of the translation components and slerp of the quaternions.
+This logic is already implemented within `TransformInterpolationBuffer`.
```
wohe: As mentioned before this is not yet implemented: since we interpolate between two trajectory nodes which are part of the optimization, we have to be able to compute derivatives, e.g. by having an automatic differentiable implementation of interpolation. This does not exist yet.

pifon2a: Done.

ojura:
I have some comments regarding landmarks:
- I think we should use numeric IDs as in rest of Cartographer. In case you have lots of observations, I think that having a full string in each of them is wasteful.
- Also for the case where you have lots of observations, perhaps introducing a sampling setting could be useful. (Perhaps this belongs in `cartographer_ros`.)
- Another useful thing would be to support landmark observations from a fixed (known) arm from the tracking frame. This would require some further adjustments to the optimization problem - for example, if you only have a positional measurement of the landmark, the orientation of the tracking frame with respect to the position of the landmark observation frame would be unknown. However, the optimizer can infer it from the direction of the whole trajectory.

pifon2a:
Hi Juraj,

1. String IDs are only keys of `std::map<string, LandmarkNode>` in pose_graph, they are not stored for every observation inside `LandmarkNode`. Having many observations for a relatively small number of landmarks won't be wasteful. 

2. There will be a sampling rate, sure. 

3. True, we have to add a possibility to specify what measurements are missing (translation-only observations, etc). I think this is related: there will be an RFC about adding fixed frame landmark observations and also relative positions from one landmark to another (think CAD models, where one can compute distances from one landmark to another and then use it as an input for cartographer).

ojura:
Thanks you, @pifon2a. I have a couple of further suggestions which I think will make Landmarks a really powerful tool, and I believe go in the direction you envisioned. I think that we will even be able to throw out fixed frame poses, since fixed frames will be just a special case of landmarks. (I find the current implementation of fixed frame poses lacking - especially due to the fact that it interpolates between the pose measurements, instead of interpolating between trajectory nodes like you do here.).

1) Add the ability to make the observations landmark-sided (_the landmark observes the robot_).
E.g. for 3D, just swap the order of `landmark_rotation` and `interpolated_pose_rotation`:

   ```c++
   ComputeUnscaledError(landmark_to_tracking_transform_,
       landmark_rotation, landmark_translation,
       interpolated_pose_rotation.data(),
       interpolated_pose_translation.data()),
   ```
    This could be controlled by a boolean inside the observation, which would specify the observation type.

2) Introduce "fixed" landmarks which can be tied to the world origin, or any world coordinate desired by the library user. This can be achieved by excluding them from optimization.

I can prepare PRs for this, if you agree.

Also left you some additional comments in https://github.com/googlecartographer/cartographer/pull/884.

ojura:
@pifon2a another minor nit :) Should `landmark_to_tracking_transform` actually be called `tracking_to_landmark_transform`? (The question is related to your implementation, not related to my PR where I swap them around). If I am correct, the `Rigid3d` contained in `landmark_to_tracking_transform` (and not its inverse!) can be chained like this:

`map_to_tracking * landmark_to_tracking`

and that would yield `map_to_landmark`, if I am correct (you would call it `landmark_to_map`?). Therefore, I think it's better to call this transform `tracking_to_landmark_transform`, so the product above looks like this:

`map_to_tracking * tracking_to_landmark` yields `map_to_landmark`.

pifon2a:
Hi Juraj,

I think `landmark_to_tracking_transform` is a correct name. This transform is chained like that:

closest_node.global_pose * observation.landmark_to_tracking_transform;
[link to code](https://github.com/googlecartographer/cartographer/blob/f606d4b91c605b7b239655de703c36d76b2ca83a/cartographer/mapping/3d/pose_graph/optimization_problem_3d.cc#L115)

`global_pose` transforms tracking to global according to the [doc](https://google-cartographer.readthedocs.io/en/latest/terminology.html).

Please, correct me if I am wrong. And thanks for looking into this!

ojura:
Ah, yes, you're right, it's me who got the transform naming convention backwards. A good example is 

https://github.com/googlecartographer/cartographer_ros/blob/15bf0db270da1a69c4543747da4230d0b0d4d13e/cartographer_ros/cartographer_ros/node.cc#L221-L222

Wouldn't it be nicer though if transforms were named in the reverse order, though? So we could have

`map_to_tracking = map_to_local * local_to_tracking`, which seems much nicer to me.

pifon2a:
I cannot agree more. The current naming is difficult to read (and to write too). Having it in this form would be much nicer. Perhaps, that's the next clean-up that has to be done: some of the transform have meaningless names, some of them are hard to read.

pifon2a:
Or, actually, we can modify Rigid3D and introduce an new member function:
landmark_to_tracking_transform.Transform(tracking_to_global).Transform(global_to_whatever).

Or it can be an operator ().

SirVer:
Inside lyft, we use `A_from_B` for transforms, which makes it as nice to read as ojura is suggesting, without needing to touch the math and convert to multiplication from left.

pifon2a:
@SirVer : nice, this one makes more sense. Created an [issue](https://github.com/googlecartographer/cartographer/issues/940).

PR #19, opened 2018-01-22
Serialization header RFC

ojura:
[Rendered](https://github.com/larics/googlecartographer_rfcs/blob/serialization-header/text/0000-serialization-header.md)


drigz:
+1 for proto stream format version - this would be so much nicer than the current failure mode (`terminate called after throwing an instance of 'std::out_of_range` in the most recent case)

cschuet (approved):

cschuet (approved):

PR #21, opened 2018-01-30
Add service response RFC.

MichaelGrupp:
[Rendered version](https://github.com/MichaelGrupp/rfcs/blob/services/text/0013-improve-ros-service-responses.md)


MichaelGrupp:
@ojura Follow-up to our discussion.

ojura:
You nailed everything. Always return true and fill out a more detailed result field inside the response.

 I'd like to amend your dictionary approach: you can define enum-like constants inside the message, [see here](https://answers.ros.org/question/9427/enum-in-msg/?answer=105806#post-id-105806). I think that would be ideal, and addresses your dictionary divergence concern, since you can use the constants defined in the message.

MichaelGrupp:
Added your suggestion.

However, I don't know if there could be overlap with the gRPC stuff that @cschuet developed? Such that we might not want to maintain separate error codes for both protocols.

cschuet:
For gRPC services we will follow the standard [gRPC error codes schema](https://developers.google.com/maps-booking/reference/grpc-api/status_codes). I would be also fine with error code + detailed string message. Maybe that'd be even more convenient. I think if you need to invent your own error code schema my suggestion would be to use the gRPC schema.

MichaelGrupp:
Ok, I'm also in for keeping it as simple as possible.

The gRPC codes look simple and generic and still flexible enough that we can use them for potential new services. And we can assume that their definition won't change anytime soon.

So a template `.srv` could be:
```
# request params
---
uint8 status_code  # 0..15 as in gRPC definition
string status_msg
# other response params
```

This way we can simply refer to the gRPC definition in the documentation if someone is interested in implementing a service caller with automatic error handling.

I guess for most users the status message has the most value anyway.

ojura:
How about a generic e.g. `ServiceStatus` message, which would include `uint code`, the status code uint constants 0..15, and a `string msg`. And each service response has a single field named e.g. `ServiceStatus status`.

MichaelGrupp:
To be honest, I don't mind :) Since the receiver should know the `.srv` definition anyway, I don't see a real advantage of further abstraction... but also no disadvantage.

ojura:
The advantage is that you have everything defined in a single place, especially the grpc-like status constants, so there’s no duplication when you use them for multiple services. I think that beats using a plain status code integer.

cschuet (approved):
Do you want to make any more changes? Otherwise I think this is ready to merge.

MichaelGrupp:
Added @ojura's suggestion. Other than that, it's ready to be merged.

cschuet:
@MichaelGrupp @ojura Thanks for contributing!

MichaelGrupp:
PR: https://github.com/googlecartographer/cartographer_ros/pull/708

PR #23, opened 2018-02-08
Write RFC for monitoring.

stevewolter:
[Rendered version](https://github.com/stevewolter/rfcs/blob/master/text/0014-monitoring.md)
[Prometheus overview](https://prometheus.io/docs/introduction/overview/)

### DIFF
--- a/text/0014-monitoring.md
+++ b/text/0014-monitoring.md
@@ -1,0 +1,90 @@
+# Collect and Expose Metrics
+
+## Summary
+
+[summary]: #summary
+
+The goal is to collect and expose structured metrics about the quality and
+status of the SLAM algorithm and the server that runs it.
+
+## Motivation
+
+[motivation]: #motivation
+
+### Dashboard for humans
+
+Human developers debug SLAM performance, and human operators need to monitor key
+metrics for the robot. Both can profit greatly from a nice dashboard that shows
+key cartographer metrics, such as a histogram of constraint quality.
+
+Today, developers often find themselves searching in log output, and the only
+way to monitor Cartographer is to run rviz at runtime. We need to provide a more
+structured way that also works for multiple instances or longer time scales.
+
+### Reference signal for robot control
+
+Some manufacturers would like to have an observable for SLAM performance so that
+they can make downstream decisions due to bad SLAM. Collected metrics could
+provide these reference signals with little intrusion in Cartographer code.
+
+## Approach
+
+[approach]: #approach
+
+### Cartographer facade
+
+Inside the cartographer library, we establish a system-agnostic facade for
+collecting metrics. The facade is compatible with all important monitoring
+systems (Prometheus, Stackdriver, OpenCensus), but doesn't introduce build
+dependencies into the core cartographer lib. The facade lives in the
+cartographer::metrics namespace and includes these core classes:
+
+*   `Counter`: A single, always-increasing count of events, e.g. "number of loop
+    iterations".
+*   `Gauge`: A value that changes both up and down, e.g. "memory used" or "open
+    requests".
+*   `Histogram`: A thread-safe histogram with predefined buckets and constant
+    memory usage. It's different from `cartographer::common::Histogram`, which
+    is thread-unsafe and does the bucketing after value collection. Over time, I
+    think that there would be potential to merge the two Histogram classes.
+*   `CounterFamily`, `GaugeFamily`, `HistogramFamily`: An interface for
+    obtaining `Counter`/`Gauge`/`Histogram` instances that share a common
+    purpose and only differ by a set of labels.
+*   `Registry`: An interface that yields `HistogramFamily`, `CounterFamily` and
+    `GaugeFamily`. The prototype called this `FamilyFactory`, but `Registry`
+    should be clearer and less Prometheus-specific.
+
+Typically, classes that support monitoring would offer a static function
+`RegisterMetrics` method that takes a reference to the Registry and stores
+references to `Counter`, `Gauge` and `Histogram` in static file-global
+variables. A single function `cartographer::RegisterMetrics` calls all the
+individual `RegisterMetrics` functions. To avoid race conditions, we require
+that `RegisterMetrics` must be called at startup time, before any multithreading
+is launched.
+
+Cartographer code initializes the static file-global variables with fake
+instances of `Counter`, `Gauge` and `Histogram` that don't record any data. Code
+that records metrics can be written agnostic of the attachment of any monitoring
+system.
+
+### Implementing the facade
+
+`cartographer_grpc` contains a concrete implementation of the facade using the
+[Prometheus C++ client library](https://github.com/jupp0r/prometheus-cpp). An
+additional dependency to it is introduced in `cartographer_grpc` behind the
+optional flag `BUILD_GRPC`. We choose [Prometheus](https://prometheus.io/)
+because it is a stable, well-established monitoring system for custom metrics
+whose network protocol is also supported by other monitoring systems (e.g.
+Stackdriver). Prometheus follows the best practices for time series data
+outlined in [Practical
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
Just a small formality: please put a link to the rendered version the PR description: 
`[Rendered version](https://github.com/stevewolter/rfcs/blob/master/text/0014-monitoring.md)`

 and please also change to stick to this guideline (see README):
> Make sure to have one sentence per line in the RFC as it makes reading diffs much easier. There is no characters per line limit for RFCs.

MichaelGrupp:
Other than that: 👍

stevewolter:
Thanks for the (positive) feedback, appreciate it. Reformatting done, link added.

MichaelGrupp:
Besides counters, gauges, histograms... how about an (boolean) `Indicator`?

stevewolter:
Essentially, this would be a gauge of bool type? I like the idea of typing gauges, even if neither Prometheus nor OpenCensus support that. (But Google-internal Monarch does, so there's hope.) I'll update the PR on Monday.

stevewolter:
IIUC, this proposal should get some outside-of-Google review. Holger, you've expressed some interest, could you PTAL or suggest a reviewer? I've added a link to the [Prometheus overview[(https://prometheus.io/docs/introduction/overview/), if you want more background.

text/0014-monitoring.md
```diff
@@ -0,0 +1,66 @@
...
+
+Inside the cartographer library, we establish a system-agnostic facade for collecting metrics.
+The facade is compatible with all important monitoring systems (Prometheus, Stackdriver, OpenCensus), but doesn't introduce build dependencies into the core cartographer lib.
+The facade lives in the cartographer::metrics namespace and includes these core classes:
+
+*   `Counter`: A single, always-increasing count of events, e.g. "number of loop iterations".
+*   `Gauge`: A value that changes both up and down, e.g. "memory used" or "open requests".
+*   `Histogram`: A thread-safe histogram with predefined buckets and constant memory usage.
+    It's different from `cartographer::common::Histogram`, which is thread-unsafe and does the bucketing after value collection.
+    Over time, I think that there would be potential to merge the two Histogram classes.
```
MichaelGrupp, line 42: AFAIK, the `cartographer::common::Histogram`s are currently only used for debug logging (dumping a string representation). If the new histograms support a `ToString()` method, we should get rid of the old ones.

text/0014-monitoring.md
```diff
@@ -0,0 +1,66 @@
...
+*   `CounterFamily`, `GaugeFamily`, `HistogramFamily`: An interface for obtaining `Counter`/`Gauge`/`Histogram` instances that share a common purpose and only differ by a set of labels.
```
MichaelGrupp: The purpose of the "families" is not completely clear to me from this sentence. Looking at the [concrete example](https://github.com/stevewolter/cartographer/blob/6985ed9c68020986247afa211780cca39baaa3b7/cartographer/mapping_3d/pose_graph/constraint_builder.cc#L46) in your prototype they are used as a sort of collection for the three constraint builder histograms. Why should they only differ by the labels, and not e.g. in the boundaries?

MichaelGrupp:
I can confirm that monitoring is a very important topic for us at Magazino as we already collect various performance statistics and status of the robot fleets in databases. We're not using Prometheus for visualization, but I'm really in favor of having generic monitoring interfaces for SLAM.

Concrete use cases would be live visualization / debugging / watchdogs but also long-term data analysis coupled with other components (e.g. queries like localization quality over time/mileage/environment changes etc.).

---

Regarding the `Indicator`: boolean gauge sounds a bit whacky to me. What I had in mind with this is not really a metric (i.e. measuring a quantity) but more some sort of simple status information that could be useful for monitoring the state from user space (e.g. "optimization running" or errors).

Regarding types: might be overkill, but templates?

text/0014-monitoring.md
stevewolter: That will be a bit of implementation work, but sounds like a good option. In practice, I think we're mostly looking at a slightly different usage pattern - monitoring histograms are per definition process-globals, while the current histograms seem to be constructed more ad-hoc.

Would you mind if we left the current hand-wave in place and dealt with it at a later point?

text/0014-monitoring.md
stevewolter: Yes, I also find it difficult to put the distinction into words. May I just throw the long description at you and you point out what you find important from a cartographer dev's view, and if it is clearer?

A family is essentially an aggregation domain. All metrics in the same family are "about the same thing" and can be aggregated safely. For example, a family might be "RPC  responses", with individual labels calling out RPC methods and RPC result classes. It's totally reasonable to just sum over all the labels in the family to get the total number of RPCs. For metrics in different families, this is generally not true - it's apples and oranges. To enable the relationship, the family shares the same type and the same metric name.

Histograms are the most complex case here, because of the question of boundaries. Histograms are an afterthought in varz (Prometheus' ancestor), and you still notice. Technically, you could set the boundaries differently for each individual histogram, but then you lose the easy aggregation property :-(.

stevewolter:
Thanks for pointing out the context and importance, Michael.

Re Indicator: "Gauge" in the monitoring context mostly means that this is a thing that's going both up and down, as opposed to a counter. If I understand the context right, this would still expose internal cartographer state, and just be a much simpler concept of "state"?

Re types: Depends strongly on Cartographer's coding style. There is a bit of duplication between the various types, and for example the families could be easily templated. On the other hand, we're looking at maybe 5 instantiations. At Google, this would be on the borderline to frivolous use of templating, but I realize that we're more template-averse than most shops. I don't have strong feelings there. What's your feeling?

text/0014-monitoring.md
MichaelGrupp: Ahh thanks for the long description, I think I get the point now. To put it in simple words how I see it: a family is an aggregate of metrics that "makes sense", i.e. only contains instances that can be compared / summed / etc because they share the same (e.g. mathematical) properties.

However, when I look at your example code: isn't putting the `rotational_score` and the `low_resolution_score` into the same family a violation of this pattern? Although I'm not so familiar with the 3D constraint builder, they sound like different metrics. At least *inside* the family, we would be dealing with apples and oranges.

text/0014-monitoring.md
MichaelGrupp: Sure.

MichaelGrupp:
Yes, the purpose of my idea is to facilitate analysis of the internal Cartographer state from outside. This is currently not very straightforward sometimes. But for now I'd say maybe the status thing is a bit out of the scope of this particular RFC, which deals more with metrics.

My view on templates is similar, I'm not a huge fan but I think that the templated histograms, gauges and counters are a good practice to facilitate dealing with different numeric types (int, uint, float, double...).

SirVer:
I am sorry, I cannot provide a lot of insight into this topic right now. Unassigned myself.

text/0014-monitoring.md
gaschler: cartographer::common::Histogram stores all data into an ever-growing array and computes bucket ranges.
cartographer::metrics will not store all data but increments to pre-defined bucket counts, hidden behind a facade, so it behaves quite differently.

text/0014-monitoring.md
gaschler: Boundaries need to match so they can be aggregated.
As an example, fixed-size bucket counts of infra and inter constraint scores can be aggregated to constraint scores (if the bucket ranges match).

gaschler: You could add your explanation from the comment "Metrics in the same family can safely be aggregated."

gaschler (approved):

text/0014-monitoring.md
stevewolter: Done.

PR #24, opened 2018-02-08
Add a pause/continue service to cartographer_ros

MichaelGrupp:
[Rendered version](https://github.com/MichaelGrupp/rfcs/blob/pause-service/text/0000-pause-service.md)


gaschler:
As I mentioned offline, the idea of the feature sounds good to me if there is an urgent need by the community.
However, I am afraid that it entails large changes to Collator, PoseExtrapolator and generally local SLAM because cartographer does not expect jumps in time.
So in order to move forward, more detailed plans how to implement this would be needed.

MichaelGrupp:
Moved to backlog.

PR #25, opened 2018-02-17
Add code structure RFC.

pifon2a:
[rendered version](https://github.com/pifon2a/rfcs/blob/e11bca586f8cc733063ea9c94bdade71086ed9ea/text/0000-code-structure.md)


text/0000-code-structure.md
```diff
@@ -0,0 +1,83 @@
...
+  3d/
+sensor/
+transform/
+```
+
+## Discussion Points
+[discussion]: #discussion
+[Internal Headers RFC](https://github.com/pifon2a/rfcs/blob/master/text/0003-internal-headers.md) talks about moving internal implementation of gRPC code to `cartographer/grpc/internal`.
+Shouldn't it be `cartographer/internal/grpc`?
+If no, then why do we have `cartographer/internal/mapping*` and not `cartographer/mapping*/internal`?
```
gaschler: grpc/internal is slightly easier to combine with build flags because you enable or disable a folder.
Other than that, as long as it is consistent I'm ok with both.

gaschler (approved):

PR #26, opened 2018-02-20
Add a monitoring bridge and interface to cartographer_ros

MichaelGrupp:
[Rendered version](https://github.com/MichaelGrupp/rfcs/blob/monitoring-ros/text/0000-monitoring-ros.md)

As discussed, we want to take care of this functionality and will do some prototyping once the core monitoring features are ready to use. This PR will be updated accordingly.


MichaelGrupp:
Implementation of internal components:
An (example) pull request with 3rd-party-dependency-free Gauge, Counter and Histogram implementations and a FamilyFactory without exposer: https://github.com/googlecartographer/cartographer_ros/pull/906

The registration in node_main.cc is purely for demo purposes, the registry exposes nothing at the moment because the ROS interface is not implemented. But the metrics do receive values.

text/0000-monitoring-ros.md
```diff
@@ -0,0 +1,46 @@
...
+### Public
+
+* Add suitable message formats for the metrics to `cartographer_ros_msgs/msg/metrics` (to be defined, WIP).
+
+* Add a service `/collect_metrics` that collects the recent metrics from the registry in its callback and replies with a metrics message.
+
+## Discussion Points
+[discussion]: #discussion
+
+Would probably require custom gauge, counter, and histogram implementations? E.g. to support increment/decrement of gauge as done via Prometheus [here in cartographer's implementation](https://github.com/googlecartographer/cartographer/blob/master/cartographer/cloud/metrics/prometheus/family_factory.cc#L70) for example.
```
gaschler: On the Cartographer side, we require Prometheus mostly for its standardized format to expose the metrics, the actual counting is only a few lines of code.
So I guess here in this case it's nice to do the counting ourselves so cartographer_ros does not depend on Prometheus (which is not ROS-packaged).

MichaelGrupp: > the actual counting is only a few lines of code.

Take a look at the prototype code PR linked in the RFC PR, I implemented these metrics already.

text/0000-monitoring-ros.md
```diff
@@ -0,0 +1,96 @@
...
+### ROS Interface
+
+Add suitable message formats for the metrics to `cartographer_ros_msgs/msg/metrics`.
+All available metrics are aggregated in a single `Metrics.msg`, this way the service caller has access to all data and can filter by itself if needed:
+```
+std_msgs/Header header
+  uint32 seq
+  time stamp
+  string frame_id
+cartographer_ros_msgs/CounterFamily[] counter_families
```
gaschler: As I commented to the code https://github.com/googlecartographer/cartographer_ros/pull/906#pullrequestreview-131574621 perhaps there is format with fewer new msg definitions?

MichaelGrupp: Good point, I removed the details of the message format in the RFC text. If the general concept of the RFC is good to go, we can merge it and discuss these details in the code PR (as we already started ;) )

gaschler (approved):

PR #29, opened 2018-04-01
Update 0018-odometry-use-in-optimization-problem.md

ryo-robo:
Thank you for commenting how to use wheel odometry at pose graph constraint.
Could you update below four parameters' name, because there aren't these parameters?

The only parameters added for the new approach were:
```lua
POSE_GRAPH.optimization_problem.initial_pose_translation_weight = 1e5
POSE_GRAPH.optimization_problem.initial_pose_rotation_weight = 1e5
POSE_GRAPH.optimization_problem.odometry_translation_weight = 1e5
POSE_GRAPH.optimization_problem.odometry_rotation_weight = 1e1
```

### DIFF
--- a/text/0018-odometry-use-in-optimization-problem.md
+++ b/text/0018-odometry-use-in-optimization-problem.md
@@ -61,13 +61,6 @@ Current upstream, i.e. implicit odometry-based constraints:
 Proposed explicit, separate weighting:
 ![upstream](0018-assets/raw-pointcloud_separate_constraints.png)
 
-The only parameters added for the new approach were:
-```lua
-POSE_GRAPH.optimization_problem.initial_pose_translation_weight = 1e5
-POSE_GRAPH.optimization_problem.initial_pose_rotation_weight = 1e5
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
These parameters just describe what was used in the experiment with the prototype, they are not final.

The new feature is not yet available in Cartographer, it was just a proposal (RFC) until only a few days ago. But I will commit the feature described in this RFC soon, so expect it in the next few days.

The final name of the parameters will change, as discussed [here](https://github.com/googlecartographer/rfcs/pull/9#issuecomment-374663523), is that what you mean?

ryo-robo:
Thank you for your kind answer.


>The new feature is not yet available in Cartographer, it was just a proposal (RFC) until only a few days ago.

I got current situation. 


>But I will commit the feature described in this RFC soon, so expect it in the next few days.

I'm looking forward to pulling your commit ! 


Thanks in advance.

MichaelGrupp:
The actual pull request for the code is here: https://github.com/googlecartographer/cartographer/pull/1029

PR #31, opened 2018-05-08
Write GlobalSlamResultCallback RFC.

cschuet:
[rendered](https://github.com/cschuet/rfcs/blob/c9cd13ef61c4ca491c39b60b5983f674c126512a/text/0000-global-slam-callback.md)

See proposed implementation [here](https://github.com/googlecartographer/cartographer/pull/1143).

### DIFF
--- a/text/0000-global-slam-callback.md
+++ b/text/0000-global-slam-callback.md
@@ -1,0 +1,36 @@
+# Global SLAM Result Callback
+
+## Summary
+[summary]: #summary
+
+For local SLAM a `LocalSlamResultCallback` can be registered when creating new trajectories which is invoked whenever new `RangefinderData` is added to the trajectory.
+Similarly we propose to add a `GlobalSlamResultCallback` which is invoked whenever the optimization problem has been solved.
+
+## Motivation
+[motivation]: #motivation
+
+[Issue #762](https://github.com/googlecartographer/cartographer_ros/issues/762) describes bandwidth issues when publishing the `/trajectory_node_list`.
+To remedy this [Issue #1022](https://github.com/googlecartographer/cartographer/issues/1022) proposes the introduction of a mechanism to signal the completion of an optimization problem.
+The introduction of a `GlobalSlamResultCallback` would accomplish that.
+
+Moreover for cloud-based mapping we would like to the ability to stream the robots pose to the cloud.
+With the introduction of a `GlobalSlamResultCallback` this code would look much more natural as upload would occur as a result of the callback invocation.
+
+## Approach
+[approach]: #approach
+
+We propose to introduce a callback with the following signature:
+
+```
+using GlobalSlamResultCallback = std::function<void(
+    const std::map<int, SubmapId>&, const std::map<int, NodeId>&)>;
+```
+
+i.e. the callback gets passed two maps, mapping trajectory IDs to the last optimized `SubmapId` and `NodeId`.
+
+We would change the public API by adding the `GlobalSlamResultCallback` as a constructor argument to `MapBuilder` which passes it on as a const'or argument to either `PoseGraph2d` or `PoseGraph3d`.
+The callback is invoked in `PoseGraph2/3d::RunOptimization()`.
+
+## Discussion Points
+[discussion]: #discussion
+
--- a/text/0000-global-slam-callback.md
+++ b/text/0000-global-slam-callback.md
@@ -29,7 +29,7 @@ using GlobalSlamResultCallback = std::function<void(
 i.e. the callback gets passed two maps, mapping trajectory IDs to the last optimized `SubmapId` and `NodeId`.
 
 We would change the public API by adding the `GlobalSlamResultCallback` as a constructor argument to `MapBuilder` which passes it on as a const'or argument to either `PoseGraph2d` or `PoseGraph3d`.
-The callback is invoked in `PoseGraph2/3d::RunOptimization()`.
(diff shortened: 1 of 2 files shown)


cschuet:
@ojura PTAL

ojura:
Thanks for writing the RFC. Here are some thoughts.

- `LocalSlamResultCallback` returns a `LocalSlamResult`, which includes the local pose, the node ID, and the point cloud used for scan matching. In other words, the complete package - the callback user has been passed everything they could possibly need related to the local SLAM result. Shouldn't an analogous `GlobalSlamResult` include all optimized trajectory node and submap (and landmark node) poses, as they really are a result of optimization? 

  However, computing these is not a simple copy operation, and the user might not need this. So it might be better to leave it to the user to call `GetTrajectoryNodePoses()` or whatever they're interested in, if they need it -- just like in your proposal. So our `GlobalSlamResultCallback` is not actually passing a global SLAM _result_, only a notification that optimization has occured. Perhaps a better fitting name would be `GlobalSlamOptimizationCallback`.

- > Moreover for cloud-based mapping we would like to the ability to stream the robots pose to the cloud. With the introduction of a GlobalSlamResultCallback this code would look much more natural as upload would occur as a result of the callback invocation.

  This is not exactly talking about the scenario where the cloud perfoms global SLAM, right? In that case, wouldn't we be _downloading_ optimized poses from the cloud back to the robots?

text/0000-global-slam-callback.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+using GlobalSlamResultCallback = std::function<void(
+    const std::map<int, SubmapId>&, const std::map<int, NodeId>&)>;
+```
+
+i.e. the callback gets passed two maps, mapping trajectory IDs to the last optimized `SubmapId` and `NodeId`.
+
+We would change the public API by adding the `GlobalSlamResultCallback` as a constructor argument to `MapBuilder` which passes it on as a const'or argument to either `PoseGraph2d` or `PoseGraph3d`.
+The callback is invoked in `PoseGraph2/3d::RunOptimization()`.
```
gaschler: I think in `HandleWorkQueue` after `RunOptimization` and before taking the mutex for trimming is better. (Note here that the callback happens before trimming.)

text/0000-global-slam-callback.md
```diff
@@ -0,0 +1,36 @@
...
+With the introduction of a `GlobalSlamResultCallback` this code would look much more natural as upload would occur as a result of the callback invocation.
+
+## Approach
+[approach]: #approach
+
+We propose to introduce a callback with the following signature:
+
+```
+using GlobalSlamResultCallback = std::function<void(
+    const std::map<int, SubmapId>&, const std::map<int, NodeId>&)>;
```
gaschler: Add `/* trajectory_id */`

gaschler (approved):
Thanks

text/0000-global-slam-callback.md
ojura: Suppose someone, somewhere calls `RunOptimization` directly. Is it their responsibility to invoke the callback as well?

cschuet:
@ojura Thanks for your comments.

> Thanks for writing the RFC. Here are some thoughts.
> 
> LocalSlamResultCallback returns a LocalSlamResult, which includes the local pose, the node ID, and the point cloud used for scan matching. In other words, the complete package - the callback user has been passed everything they could possibly need related to the local SLAM result. Shouldn't an analogous GlobalSlamResult include all optimized trajectory node and submap (and landmark node) poses, as they really are a result of optimization?
> 
> However, computing these is not a simple copy operation, and the user might not need this. So it might be better to leave it to the user to call GetTrajectoryNodePoses() or whatever they're interested in, if they need it -- just like in your proposal. So our GlobalSlamResultCallback is not actually passing a global SLAM result, only a notification that optimization has occured. Perhaps a better fitting name would be GlobalSlamOptimizationCallback.

Renaming sounds good. I was also considering passing the whole optimized data to make it more similar to the LocalSlamResultCallback, but decided against that for exactly the reasons you outlined above. A more puristic approach would be to simply have a std::function<void()> but the subscriber having access to the last optimized node / submap information turned out to be useful for the use-cases I sketched.

> 
>> Moreover for cloud-based mapping we would like to the ability to stream the robots pose to the cloud. With the introduction of a GlobalSlamResultCallback this code would look much more natural as upload would occur as a result of the callback invocation.
> 
> This is not exactly talking about the scenario where the cloud perfoms global SLAM, right? In that case, wouldn't we be downloading optimized poses from the cloud back to the robots?

I am mainly interested in low-latency fleet visualization. Whenever new local and global poses become available I would want to upload those into the cloud. This is different from streaming IMU, odometry and submaps to the cloud and where they need to make it into through the optimization problem first.

I would explicitly stream base_link => local  and local => global for visualization, so an explicit global SLAM callback is convenient for that.

ojura:
>I am mainly interested in low-latency fleet visualization. Whenever new local and global poses become available I would want to upload those into the cloud. This is different from streaming IMU, odometry and submaps to the cloud and where they need to make it into through the optimization problem first.
>
> I would explicitly stream base_link => local and local => global for visualization, so an explicit global SLAM callback is convenient for that.

Okay, but isn't local => global computed on the cloud, as it's doing global SLAM? Sorry for being off-topic.

cschuet:
@ojura 

> Okay, but isn't local => global computed on the cloud, as it's doing global SLAM? Sorry for being off-topic.

It is also calculated in the cloud. On the client however we usually run Cartographer in pure_localization mode (which is also global SLAM in some sense) and that gives us a medium-latency optimized pose of the robot with respect to its frozen localization map.

PR #32, opened 2018-05-24
Serialization format for 1.0

sebastianklose:
This RFC proposes a new versioned file format for serialized `PoseGraphs`.

[Rendered version](https://github.com/sebastianklose/rfcs/blob/serialization_format/text/0021-serialization-format.md)


text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
+# Serialization Format
+
+## Summary
+
+[summary]: #summary
+
+Definition of a versioned, backwards-compatible and extensible serialization format for cartographer-maps.
+
```
pifon2a: nit: s/cartographer-maps/cartographer state

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+This RFC is a small deviation of the previously proposed header format from [RFC-0015](https://gitub.com/googlecartographer/rfc/text/0015-serialization-header.md).
+
+## Motivation
+
+[motivation]: #motivation
+
+At the moment, the file format for serialized *PoseGraphs* is defined purely within the serialization & deserialization routines of cartographer. Since we serialize all information into a binary stream of `protobuf` messages (*pbstream*) without a format-version information, extending or changing the format inevitably leads to a breaking change for cartographer users, as there is no way of falling back to load previously serialized maps.
```
pifon2a: nit: PoseGraph (without 's')

pifon2a: %s/cartographer/Cartographer/g

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+* Additionally we will provide a migration tool to re-write already serialized pbstreams into the new format.
+
+## Discussion Points
+
+[discussion]: #discussion
+* Version string for new format: 0.9 or 1.0?
+* Duplicating protos to separate storage and *wire* format as of protobuf-best-practices
+* Naming preferences:
+   	- `SerializationHeader`
+   	- `SerializedPoseGraphData`
```
pifon2a: I would leave like it was: `SerializedData`. `SerializedPoseGraphData` is way too long, i think.

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+
+For being able to identify the underlying file-format, we introduce a `SerializationHeader` message.
+
+```
+message SerializationHeader {
+  string format_version = 1;
+}
+```
+
+We further introduce a `SerializedPoseGraphData` protobuf message, which is meant to be able to contain `one-of` any of the message types required for serializing the mapping state:
```
pifon2a: nit: `oneof`

text/0000-serialization-format.md
pifon2a: I am not sure that it is completely correct to say that we have a format for "serialized PoseGraph". We are trying to serialize the whole state of Cartographer. Now it is mostly the pose graph and trajectory/sensor options/data.

pifon2a (commented):
Awesome! just nits.

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
+# Serialization Format
+
+## Summary
+
+[summary]: #summary
+
+Definition of a versioned, backwards-compatible and extensible serialization format for cartographer-maps.
+
+This RFC is a small deviation of the previously proposed header format from [RFC-0015](https://gitub.com/googlecartographer/rfc/text/0015-serialization-header.md).
```
pifon2a: the link is broken because of the typo: "gitub"

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+
+At the moment, the file format for serialized *PoseGraphs* is defined purely within the serialization & deserialization routines of cartographer. Since we serialize all information into a binary stream of `protobuf` messages (*pbstream*) without a format-version information, extending or changing the format inevitably leads to a breaking change for cartographer users, as there is no way of falling back to load previously serialized maps.
+
+This RFC is going to address this issue by introducing a `SerializationHeader` message and introducing an interface for parsing serialized *PoseGraphs*.
+
+## Approach
+
+[approach]: #approach
+
+For being able to identify the underlying file-format, we introduce a `SerializationHeader` message.
```
gaschler: "file-format"->"file format"

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+| *Generic File Format* |
+| :---: |
+| SerializationHeader |
+| (SerializedPoseGraphData)* |
+
+### Version 1.0 file-format
+
+Using the definitions above, the proposed file-format of version 1.0 is as follows
+
+|*Version 1.0 fileformat (proto order)*|
```
gaschler: Is this sequence strict? Will we open files where ImuData comes after LandmarkData? Files where PoseGraph is not the first SerializedPoseGraphData packet?

text/0000-serialization-format.md
sebastianklose: done

text/0000-serialization-format.md
```diff
@@ -0,0 +1,95 @@
...
+
+* To gracefully transition to the new format, we will provide an internal fallback to the *old* parsing of PbStreams. 
+Meaning, we will first try to read the a `SerializationHeader` from the pbstream and if this fails, seek back in the stream and try parsing the *old* format. This way we can at least provide functionality for parsing the last pre-header version.
+* To discourage usage of the old format, the code for serializing the mapping state will only support the new format.
+* Additionally we will provide a migration tool to re-write already serialized pbstreams into the new format.
+
+## Discussion Points
+
+[discussion]: #discussion
+* Version string for new format: 0.9 or 1.0?
```
gaschler: I'd just go with an integer 1 and count up independent of cartographer's version string.

text/0000-serialization-format.md
sebastianklose: done

text/0000-serialization-format.md
sebastianklose: done

text/0000-serialization-format.md
sebastianklose: Updated section to mention that we will rename the "old" `SerializedData` to `LegacySerializedData` for backwards compatibility.

text/0000-serialization-format.md
sebastianklose: Rephrased.

text/0000-serialization-format.md
sebastianklose: fixed.

text/0000-serialization-format.md
```diff
@@ -0,0 +1,116 @@
...
+
+By keeping all the payload related data wrapped in a single message type, we can
+also easily switch to a different underlying storage format in the future (e.g.
+[Riegeli](https://github.com/google/riegeli/)).
+
+### Generic File Format
+
+Using the above messages, the generic file format is defined as:
+
+*Generic File Format* |
```
gaschler: Please mention that we put magic bytes "0x7b1d1f7b5bf501db" in front and that all messages are gzipped (each message separately), as in `ProtoStreamWriter`.

gaschler (commented):
Generally agree. The magic bytes and compression should be mentioned to completely specify our file format here.

text/0000-serialization-format.md
sebastianklose: done.

text/0000-serialization-format.md
sebastianklose: We technically do not require to be too strict. But for easing our deserialization logic, we probably should at least enforce the `PoseGraph` and `AllTrajectoryBuilderOptions` to come directly after the header. The remaining parts we can probably loosen the ordering assumption.

text/0000-serialization-format.md
sebastianklose: I'd definitely also not directly link this with the Cartographer version 1-to-1. So maybe an int is fine if nobody objects.

text/0000-serialization-format.md
```diff
@@ -0,0 +1,122 @@
...
+:-------------------------------: |
+SerializationHeader               |
+PoseGraph                         |
+AllTrajectoryBuilderOptions       |
+(Submap)\*                        |
+(Node)\*                          |
+(ImuData)\*                       |
+(OdometryData)\*                  |
+(FixedFramePoseData)\*            |
+(TrajectoryData)\*                |
```
MichaelGrupp: I'd move `TrajectoryData` to `Node` and `Submap`, such that all data that describes the topology of the SLAM results is separated from the process input data (sensor, landmarks...).

MichaelGrupp (commented):
Related to my inline comment: what's the purpose of serializing sensor data etc at the moment?

text/0000-serialization-format.md
sebastianklose: Sounds reasonable. Will change.
Just as a note: The order of the payload as listed here, is exactly how we serialize at the moment.

sebastianklose:
@MichaelGrupp The sensor data is part of the serialized data already today. I agree that for a fixed map you probably won't need it. But if you would like to be able to improve the map (e.g. in a further batch processing over night), the original sensor information will be useful and sometimes even required to evaluate the residuals and jacobians.

text/0000-serialization-format.md
ojura: Yes, that was my proposal as well in RFC-0015: I was proposing to store unique protos (like `AllTrajectoryBuilderOptions`, `PoseGraph`) in the header. But I have no strong feelings, these can also go into `SerializedData`.

sebastianklose: I more or less see the header as meta information and therefore would prefer keeping any mapping state related data out of it.
Also this makes it easier to use Riegeli as a storage container later on.

MichaelGrupp (approved):

text/0000-serialization-format.md
gaschler: I agree with Michael.

gaschler (approved):

text/0000-serialization-format.md
```diff
@@ -0,0 +1,125 @@
...
+*   We will provide a migration tool to re-write already serialized pbstreams
+    into the new format.
+
+## Discussion Points
+
+[discussion]: #discussion
+
+*   Version string for new format: 0.9 or 1.0?
+    -   `uint32` - first version == 1
+*   Duplicating protos to separate storage and *wire* format following
```
wohe: I do not understand this sentence. What is it trying to say?

text/0000-serialization-format.md
```diff
@@ -0,0 +1,125 @@
...
+For being able to identify the underlying serialization format, we introduce a
+`SerializationHeader` message.
+
+```
+message SerializationHeader {
+  uint32 format_version = 1;
+}
+```
+
+We further change the current `SerializedData` protobuf message, which is meant
```
wohe: +1. The current state only exists because before the switch to a new protobuf library, oneof was not supported yet.

text/0000-serialization-format.md
```diff
@@ -0,0 +1,125 @@
...
+    FixedFramePoseData fixed_frame_pose_data = 8;
+    LandmarkData landmark_data = 9;
+  }
+}
+```
+
+~~For backwards compatibility, we will rename the original `SerializedData`
+definition to `LegacySerializedData`.~~ (see discussion)
+
+By keeping all the payload related data wrapped in a single message type, we can
```
wohe: +1. I think riegeli did not exist when we added the pbstream format. Would be great to remove the pbstream code and use this library instead. This would also allow us to change to faster compression.

wohe (approved):

wohe:
@MichaelGrupp @sebastianklose The original sensor data is not included, only the small part that is used in the pose graph optimization. The purpose is that you can load serialized data and combine it with new data in a newly optimized map as Michael said. This would not be possible otherwise, i.e. this data is required in this case.

text/0000-serialization-format.md
sebastianklose: I only wanted to mention this as I have read about this being a best practice.
It seems the advice is to always have two pairs of protos: One for what you use for storage and one which you use e.g. in RPC calls or also within your running system in objects. This allows you to change your storage format without the rest of your code. Right now it looks like we use the same for both.

I also already discussed with @cschuet that we should probably postpone and discuss this separately if it comes up again. I'll remove if from this RFC to avoid confusion.

PR #34, opened 2018-06-07
Replace frozen trajectory from file

gaschler:
[Rendered version](https://github.com/gaschler/rfcs/blob/delete-load/text/0000-delete-load.md)


pifon2a (approved):

gaschler:
You can find a work-in-progress implementation in https://github.com/googlecartographer/cartographer/pull/1195

cschuet:
@gaschler Thanks!

cschuet (approved):

MichaelGrupp:
Is the new state loaded concurrently? Loading a large pbstream can take its time.

gaschler:
@MichaelGrupp 
LoadStateFromFile will initially work in the handler thread, so it blocks one of the multiple handler threads.
I will add a TODO to work in a new background or "i/o" thread so we never block SLAM even if more than ~4 (= number of handler threads) concurrent calls happen.
The stub call of LoadStateFromFile returns after all data has been passed to PoseGraph. (This doesn't mean that the map is used, this the global SLAM callback will indicate.)

text/0000-delete-load.md
```diff
@@ -0,0 +1,48 @@
...
+We propose to add `PoseGraphInterface::DeleteTrajectory`, `PoseGraphInterface::GetTrajectoryStates`, and `MapBuilderInterface::LoadStateFromFile` because these functions are missing to replace a frozen trajectory ("map") at run-time.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, it is not possible to replace a finished (and frozen) trajectory at run-time.
+In fact, the trimmers are the only components that even delete parts of the pose graph.
+When we add these two functions to the public API, the use case of swapping a frozen map at run-time is possible.
+A caller can load a new frozen map, wait for global SLAM to process it (and re-localize in the new map), and delete the old frozen map.
+Another use case is to clean up old trajectories on a device that runs continuously and sometimes starts and finishes trajectories.
```
MichaelGrupp: I think the deletion can be even implicit for trajectories that have been finished and trimmed completely. Related to https://github.com/googlecartographer/cartographer/issues/1111

text/0000-delete-load.md
```diff
@@ -0,0 +1,48 @@
...
+
+We propose to add `PoseGraphInterface::DeleteTrajectory`, `PoseGraphInterface::GetTrajectoryStates`, and `MapBuilderInterface::LoadStateFromFile` because these functions are missing to replace a frozen trajectory ("map") at run-time.
+
+## Motivation
+[motivation]: #motivation
+
+Currently, it is not possible to replace a finished (and frozen) trajectory at run-time.
+In fact, the trimmers are the only components that even delete parts of the pose graph.
+When we add these two functions to the public API, the use case of swapping a frozen map at run-time is possible.
+A caller can load a new frozen map, wait for global SLAM to process it (and re-localize in the new map), and delete the old frozen map.
```
MichaelGrupp: How do we deal with the possibly different origins of the frozen map coordinate frame?

MichaelGrupp (commented):
Added some comments, also in the WIP pull request.

text/0000-delete-load.md
gaschler: Even when the coordinate frame is correct, there is absolutely no guarantee that re-localization will happen.
Is it a very frequent use case that the replacement frozen map has a different coordinate frame?
In my opinion, this mostly adds complexity, and it is already the caller's responsibility that the replacement trajectory is so similar that localization will be fine, so it'd keep it simple for now and not support different origins.

text/0000-delete-load.md
gaschler: I'm not sure if implicitly deleting is the right way to fix #1111.
Should we support `SerializeState` for a selection of trajectories in a future RFC?

MichaelGrupp: Yes improving the (selective) serialization would be something separate, this just came to my mind again when I read it.

text/0000-delete-load.md
MichaelGrupp: This also makes more sense to me if we think e.g. of the use case of streaming local map updates to a server instance and regularly updating the localization maps on the clients by a new pbstream file. In this case we'd aim to have the old and the new frozen map aligned to each other.

However, there _might_ be situations where it might be problematic, that's why I asked.

MichaelGrupp (approved):

PR #37, opened 2018-08-27
Add RFC for Docker Hub deployment.

MichaelGrupp:
[Rendered version](https://github.com/MichaelGrupp/rfcs/blob/docker-push/text/0000-use-docker-hub.md)


MichaelGrupp:
From open house discussion: consider only a base image for 3rdparty dependencies
  * because we still might want to re-build Cartographer everytime in cartographer_ros
  * not re-building protobuf, Ceres, ... can already save quite some time
  * protobuf, Ceres versions etc. are fixed anyway
  * --> updated approach accordingly

MichaelGrupp:
For reference: https://github.com/googlecartographer/cartographer_ros/pull/1021

MichaelGrupp:
Closing this for now, we have a base image for `cartographer_ros` now.

PR #38, opened 2019-01-17
Save user data in proto stream

ojura:
[Rendered](https://github.com/larics/cartographer_rfcs/blob/master/text/0000-saving-user-data-in-pbstream.md)

### DIFF
--- a/text/0000-saving-user-data-in-pbstream.md
+++ b/text/0000-saving-user-data-in-pbstream.md
@@ -25,7 +25,7 @@ Clients of libcartographer are responsible for serializing their data into the `
 
 `MapBuilder::LoadState` will return, besides the trajectory remapping like it currently does, the same vector of `UserData` protos that was passed during `SerializeState`.
 
-cartographer_ros will introduce a new proto or ROS message (up to discussion) called `NodeOptions`, which corresponds to `cartographer_ros::NodeOptions` (which currently does not have serialization).
+cartographer_ros will introduce a new proto or ROS message (up for discussion) called `NodeOptions`, which corresponds to `cartographer_ros::NodeOptions` (which currently does not have serialization).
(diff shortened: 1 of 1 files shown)


ojura:
Conclusions from the open house:

- `MapBuilderOptions` will be saved to the `AllTrajectoryOptions` proto, to be renamed as `AllOptions` or `BuilderOptions` (open to suggestions for the naming)
- the pbstream info tool can be used to recover them afterwards; no support for directly loading the embedded node options in cartographer_ros is planned (_sad face_ :) )

- options for saving other user data: 
     - users save their own protos at the beginning of the proto stream using the `Writer` API, then hand off the `Writer` to Cartographer for serialization
        - pros: users do not have to go through the entire stream to get to their data
        - cons: the produced streams will not be able to be used with upstream Cartographer, since it does not know how to skip until it hits the `Header` proto

     - users can expand the `SerializedData` proto with their own type and write out their own `SerializedData` protos using the Writer API after Cartographer is finished with serializing
       - pros: upstream Cartographer can ignore these in its huge `SerializedData` proto-type switch, no upstream changes necessary
       - cons: without interventions into Cartographer, there is no possibility to write the user data at the beginning of the proto stream, only at the end, which can be slow for large streams

     -  similarly to the description in the RFC, `SerializedData` gets an `Any` field, and users hand off a vector of `Any` protos to Cartographer so it can write them out at the beginning (after the header)
       - pros: user data is at the beginning so can be accessed fast
       - cons: upstream Cartographer needs to be taught to do this, API change required

I am of course rooting for option 3 :)

ojura:
Okay, the first part (options proto rename + storing map builder options) is in https://github.com/googlecartographer/cartographer/pull/1504. I'll wait for your feedback for the other one.

PR #39, opened 2019-03-01
Frontier detection RFC

ojura:
[Rendered](https://github.com/larics/cartographer_rfcs/blob/frontier/text/0000-frontier_detection.md)

### DIFF
--- a/text/0000-frontier_detection.md
+++ b/text/0000-frontier_detection.md
@@ -1,0 +1,38 @@
+# Frontier detection for Cartographer
+
+## Summary
+[summary]: #summary
+
+Introduce the paper on frontier detection and a possible plug-in system for Cartographer
+
+## Motivation and approach
+[motivation]: #motivation
+
+This RFC is a bit different than usual: I want to share a paper I wrote about a special frontier detection algorithm I devised for Cartographer.
+The pre-print of the paper is available on arXiv: https://arxiv.org/abs/1902.11061.
+A monorepo with an implementation (research-quality :-)) is available at https://github.com/larics/cartographer_frontier_detection, and a demo video with Deutsches museum at https://goo.gl/62zEUy.
+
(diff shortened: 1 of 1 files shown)


ojura:
More specifically about the plugins idea: ROS offers [pluginlib](http://wiki.ros.org/pluginlib) for dynamically loading plugins into ROS nodes. Rviz uses this (the Submaps plugin, for example).

If cartographer_ros offered a plugin API where we could execute the plugin code in `OnLocalSlamResult`, this would be sufficient for having a standalone implementation of frontier detection.

sebastianklose:
Thanks @ojura for your RFC. The results look very interesting.

Without having looked at your code in detail, I would vote against adding a plugin architecture to cartographer_ros for this purpose. I would be fine with the idea of opening up the `MapBuilderBridge` to become an interface that can be implemented differently for the frontier detection.
WDYT?

ojura:
Can you please elaborate more? You think this could be merged upstream?

The code for frontier detection is more or less fully contained in [`cartographer_ros/frontier_detection.cc`](https://github.com/larics/cartographer_frontier_detection/blob/master/cartographer_ros/cartographer_ros/cartographer_ros/frontier_detection.cc) / [`.h`](https://github.com/larics/cartographer_frontier_detection/blob/master/cartographer_ros/cartographer_ros/cartographer_ros/frontier_detection.h), and is called from `OnLocalSlamResult`. The code for frontier detection does require Cartographer(_ros) headers and access to Cartographer internals (specifically, the submap occupancy grid values, and their optimized poses). Adding the needed wiring to `MapBuilderBridge` sounds like a good idea in any case, but does not resolve the issue of this code being highly dependent on cartographer_ros (since it needs to run inside the cartographer_node executable if we want speedy access to submap grids). If it is not merged upstream, I see no other option than living in a fork of cartographer_ros like at the present, or as a standalone plugin making use of a plugin API.

I have a couple of other ideas for specializing robotics algorithms for Cartographer (for example, submap-level path planning), so that's why I'm rooting for a plugin API where we could write plugins which make use of Cartographer internals, but can live as standalone Catkin packages.

